PR #23, opened 2024-03-13
feat: Detect usage of deprecated jQuery.sap API

matz3:
So far, only the usage of global API such as `jQuery.sap` was detected.
However, when using `jQuery` as local variable via a dependency, the
usage of `jQuery.sap` was not detected, although all APIs within that
namespace are deprecated.


resources/overrides/jquery.sap.mobile.d.ts
```diff
@@ -11,7 +11,3 @@ interface JQueryStatic {
 	 */
 	device: object;
```
d3xter666, line 12: `JQueryStatic` from this file would be extended by the one in `resources/overrides/jquery.sap.d.ts`.

I was wondering wether we need to keep this one or shall we just merge them explicitly

matz3: Do you mean having them in one file? This is on purpose to allow easy detection (see `isSymbolOfJquerySapType` in `FileLinter.ts`)

d3xter666: To be honest, I don't mind it. We just need to be aware of the side effects when merging TS definitions. So, maybe just mentioning as a comment that this interface has another definition in `resources/overrides/jquery.sap.d.ts` and they could (eventually) be merged, so we would be aware of the eventual side effects.
This is the TS feature I'm reffering to: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

d3xter666 (dismissed):
LGTM

RandomByte (dismissed):

RandomByte (approved):

PR #33, opened 2024-03-14
feat: Analyze sap.ui.core.Lib.init() call

d3xter666:
JIRA: CPOUI5FOUNDATION-793

Adds checks whether `Lib.init()` has been called with `apiVersion: 2` property.


src/detectors/typeChecker/FileLinter.ts
```diff
@@ -55,6 +55,7 @@ export default class FileLinter {
 			// const nodeType = this.#checker.getTypeAtLocation(node);
 			this.analyzePropertyAccessExpression(node as ts.CallExpression); // Check for global
 			this.analyzeCallExpression(node as ts.CallExpression); // Check for deprecation
+			this.checkLibInitCall(node as ts.CallExpression); // Check for sap.ui.coreLib.init usages
```
matz3, line 58:
```suggestion
			this.checkLibInitCall(node as ts.CallExpression); // Check for sap/ui/core/Lib.init usages
```

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -207,6 +208,97 @@ export default class FileLinter {
...
+		}
+
+		const importedVarName = ((nodeExp.name.parent as unknown) as ts.CallExpression).expression.getText();
+		const fnArg = this.getFnArgument(node, {
+			callExpressionName: "init", // Method's name
+			callExpressionPropName: importedVarName,
+			fnArgument: 0, // Lib.init() we're interested only in the first arg
+		});
+
+		let libVersion;
```
matz3, line 272: This variable doesn't hold the actual version (number), but the node that defines it (or not even that in some cases). Maybe a better name and/or a type declaration would help a reader to understand this.

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -207,6 +208,97 @@ export default class FileLinter {
...
+
+		if (fnArg === null) { // The method is init, but nor argument has been found
```
matz3, line 274:
```suggestion
		if (fnArg === null) { // The method is init, but no argument has been found
```

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -207,6 +208,97 @@ export default class FileLinter {
...
+	}
+
+	checkLibInitCall(node: ts.CallExpression) {
+		const nodeExp = node.expression as ts.PropertyAccessExpression;
+		const {symbol} = this.#checker.getTypeAtLocation(nodeExp);
+		// TS parser uses some intermediate types that are not available as definitions.
+		// In this case SymbolObject which is a ts.Symbol + ts.Node and that's
+		// why we need these ugly type castings
+		const importDeclaration =
+			((((symbol as unknown) as ts.Node)?.parent?.parent as unknown) as ts.Symbol)?.getEscapedName() as string;
```
matz3, line 259: Is this really the only solution? So far we didn't need this in this file, so I wonder what's so special here.

src/detectors/typeChecker/FileLinter.ts
d3xter666: Actually, this is needed in order to calculate the poistion of the node. What about `nodeToHighlight`?

src/detectors/typeChecker/FileLinter.ts
d3xter666: I don't think we have such a case already.

What it does is to go up the tree to the sap.ui.define/require call and check the name of the imported module.
So far we directly access the sap.ui.define/require and didn't need to go up the tree.
This code verifies that the local `Lib.init()` call is actually the `sap.ui.core.Lib.init()`.

I'd appreciate any suggestions to improve this one

d3xter666: Done

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -211,6 +212,89 @@ export default class FileLinter {
...
+	getSymbolModuleDeclaration(symbol: ts.Symbol) {
+		let parent = symbol.valueDeclaration?.parent;
+		while (parent && !ts.isModuleDeclaration(parent)) {
+			parent = parent.parent;
+		}
+		return parent;
+	}
+
+	analyzeLibInitCall(node: ts.CallExpression) {
+		const nodeExp = (ts.isPropertyAccessExpression(node.expression) ||
```
RandomByte, line 256: Do we want to also detect and analyze cases like the following? (CC: @matz3)

```js
const myInit = Library.init;
myInit({
    apiVersion: 1
});
```

If yes, this part needs to be adapted to allow for call expressions that are not based on a Property- or ElementAccessExpression.

In any case, I think we should add a test for such a scenario (either positive or negative).

RandomByte (changes requested):

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -211,6 +212,89 @@ export default class FileLinter {
...
+			ts.isElementAccessExpression(node.expression)) && node.expression;
+		const nodeType = nodeExp && this.#checker.getTypeAtLocation(nodeExp);
+		if (!nodeType || nodeType.symbol?.getName() !== "init") {
```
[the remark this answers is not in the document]
d3xter666: I'd need to add also, `!nodeType` as the Type Definitions says it's of type `false | ts.Type`

test/fixtures/linter/rules/NoDeprecatedApi/old_library.js
```diff
@@ -0,0 +1,32 @@
+/*!
```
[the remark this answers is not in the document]
d3xter666: Thanks! Done

src/detectors/typeChecker/FileLinter.ts
d3xter666: Then we (potenitally) might have also destructors and all the implications that come with them: 
``js
const {init} = Library;
init({
    apiVersion: 1
});
// OR
const {init: myInit} = Library;
myInit({
    apiVersion: 1
});
``

RandomByte:
I benchmarked one of the recent commits from this branch against the latest commit from the `main` branch that this one is based on:

| Commit | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| d313912 (`main`) | 31.841 ± 0.415 | 31.198 | 32.597 |
| 13abc7f | 32.714 ± 0.914 | 31.066 | 34.025 |

src/detectors/typeChecker/FileLinter.ts
matz3: For now, we should only detect the simple case via `Lib.init` (current implementation).

d3xter666: I already tried the above scenarios and most of them seem to be feasible and easy to solve with couple lines of code. 
Only destructuring in `sap.ui.define` would require more efforts here.

RandomByte: At this point, `sap.ui.define` should already be transpiled to ESM import/export declarations. That's why you can ask the tsc checker for the type of `Library` and it resolves it via the import.

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -211,6 +212,89 @@ export default class FileLinter {
...
+			}
+
+			const key = prop.name.getText();
+			if (prop.initializer.kind === ts.SyntaxKind.FalseKeyword) {
+				properties[key] = {value: false, node: prop.initializer};
+			} else if (prop.initializer.kind === ts.SyntaxKind.NullKeyword) {
+				properties[key] = {value: null, node: prop.initializer};
+			} else if (ts.isObjectLiteralExpression(prop.initializer) && prop.initializer.properties) {
+				properties[key] = {value: this.extractPropsRecursive(prop.initializer), node: prop.initializer};
+			} else if ((ts.isIdentifier(prop.initializer) ||
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/detectors/typeChecker/FileLinter.ts
```diff
@@ -211,6 +212,69 @@ export default class FileLinter {
...
+	getSymbolModuleDeclaration(symbol: ts.Symbol) {
+		let parent = symbol.valueDeclaration?.parent;
+		while (parent && !ts.isModuleDeclaration(parent)) {
+			parent = parent.parent;
+		}
+		return parent;
+	}
+
+	analyzeLibInitCall(node: ts.CallExpression) {
+		const nodeExp = (
```
matz3, line 224: (style) we usually just check whether a node is of one of the expected types and stop (return statement) otherwise. This might be easier to read and consistent with the other code in this file.

RandomByte (approved):
LGTM 🚀

PR #41, opened 2024-03-20
feat(cli)!: In case of errors, exit with code 1

RandomByte:
This change alignes UI5 linter with ESLint (see [1]):
* No errors, maybe warnings: **Use exit code 0**
* At least one error: **Use exit code 1**
* Unexpected error (typically caused by an exception): **Use exit code 2**

**BREAKING CHANGE:** Depending on how UI5 linter is being used, this change
might change the behavior of that particular scenario. For example a
build script might abort further processing, if any linting errors
are detected.

To ignore exit code 1, for example in a bash script, you may use a
construct like this:
```sh
ui5lint && RC=$? || RC=$?
if [[ "$RC" = 2 ]]; then
    exit 1
fi
```

JIRA: CPOUI5FOUNDATION-823

[1]: https://eslint.org/docs/latest/use/command-line-interface#exit-codes

### DIFF
--- a/src/cli/base.ts
+++ b/src/cli/base.ts
@@ -141,6 +141,11 @@ async function handleLint(argv: ArgumentsCamelCase<LinterArg>) {
 	if (profile) {
 		await profile.stop();
 	}
+
+	if (res.some((file) => !!file.errorCount)) {
+		// At least one error is reported. Exit with non-zero exit code.
+		process.exit(1);
+	}
 }
 
 export default function base(cli: Argv) {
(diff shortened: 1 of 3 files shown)


flovogt (approved):

matz3 (changes requested):
Isn't this a breaking change?

RandomByte:
> Isn't this a breaking change?

True, done

flovogt:
Please add also the jira CPOUI5FOUNDATION-823 to the commit message

RandomByte:
> Please add also the jira item id CPOUI5FOUNDATION-823 to the commit message

Done 👍

flovogt (approved):

PR #48, opened 2024-03-25
feat: Check script tags for inline JS

d3xter666:
JIRA: CPOUI5FOUNDATION-826


src/linter/html/linter.ts
```diff
@@ -0,0 +1,43 @@
...
+	}));
+
+	jsScriptTags.forEach((tag) => {
+		const scriptContent = tag.textNodes?.map((tNode) => tNode.value).join("").trim();
+
+		if (scriptContent) {
+			report.addMessage({
+				node: tag,
+				severity: LintMessageSeverity.Error,
+				ruleId: "ui5-linter-csp-compliance",
```
d3xter666, line 32: Did we agree on `severity` and the `ruleId`?

tsconfig.base.json
```diff
@@ -14,6 +14,9 @@
 		// Allow JS in order to check the JSTokenizer
 		"allowJs": true,
 
+		// Allow parsing of renamed html (to jsx) files
+		"jsx": "preserve",
```
d3xter666, line 18: In order not to transform the inpout of the JSX file, we need to set this option to "preserve".
More information can be seen here: https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage

test/fixtures/linter/rules/CSPCompliance/NoInlineJS.html
```diff
@@ -0,0 +1,39 @@
+<html>
+
+<head>
+	<title>No inline JS</title>
```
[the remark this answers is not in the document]
RandomByte: From my discussion with FW, let's add a meta tag to see whether the TS compiler still handles the document fine.

Note how the meta tag does not close. That's valid HTML but not valid XML (hence not valid JSX? 🤔)

src/detectors/transpilers/html/parser.ts
```diff
@@ -0,0 +1,59 @@
...
+		saxWasmBuffer = await fs.readFile(saxPath);
+	}
+
+	return saxWasmBuffer;
+}
+
+async function parseHtml(contentStream: ReadStream, parseHandler: (type: SaxEventType, tag: Detail) => void) {
+	const options = {highWaterMark: 32 * 1024}; // 32k chunks
+	const saxWasmBuffer = await initSaxWasm();
+	const saxParser = new SAXParser(SaxEventType.OpenTag | SaxEventType.CloseTag, options);
```
RandomByte, line 21:
```suggestion
	const saxParser = new SAXParser(SaxEventType.CloseTag, options);
```

Since we are only interested in closing tags, do not request the OpenTag event

src/detectors/typeChecker/index.ts
```diff
@@ -212,11 +217,20 @@ export class TsProjectDetector extends ProjectBasedDetector {
...
 			// Rewrite fs-paths to virtual paths
-			resourcePaths = allResources.map((res: Resource) => {
+			resourcePaths = [...allResources, ...allTestResources].map((res: Resource) => {
 				if (absoluteFilePaths.includes(res.getSourceMetadata().fsPath)) {
 					return res.getPath();
 				}
-			}).filter(($: string | undefined) => $);
+			})
+				.filter(($: string | undefined) => $)
+				.map((res) => {
```
RandomByte, line 226: Move into the map function three lines above the reduce the number of iterations over this possibly large array

src/linter/html/HtmlReporter.ts
```diff
@@ -0,0 +1,84 @@
...
+	constructor(filePath: string) {
+		this.#filePath = filePath;
+	}
+
+	addMessage({node, message, messageDetails, severity, ruleId, fatal = undefined}: ReporterMessage) {
+		if (fatal && severity !== LintMessageSeverity.Error) {
+			throw new Error(`Reports flagged as "fatal" must be of severity "Error"`);
+		}
+
+		let line = 1, column = 1;
```
RandomByte, line 21:
```suggestion
		let line = 0, column = 0;
```

Probably? Since we always do `+ 1` later

src/linter/html/HtmlReporter.ts
```diff
@@ -0,0 +1,84 @@
...
+
+		if (messageDetails) {
+			msg.messageDetails = resolveLinks(messageDetails);
+		}
+
+		this.#messages.push(msg);
+	}
+
+	addCoverageInfo({node, message, category}: ReporterCoverageInfo) {
+		let line = 1, column = 1, endLine = 1, endColumn = 1;
```
RandomByte, line 43:
```suggestion
		let line = 0, column = 0, endLine = 0, endColumn = 0;
```

Same here?

src/linter/html/linter.ts
```diff
@@ -0,0 +1,43 @@
+import {taskStart} from "../../detectors/util/perf.js";
+import {extractScriptTags} from "../../detectors/transpilers/html/parser.js";
+import {LintMessageSeverity} from "../../detectors/AbstractDetector.js";
+import HtmlReporter from "./HtmlReporter.js";
+
+import type {TranspileResult} from "../../detectors/transpilers/AbstractTranspiler.js";
+import type {ReadStream} from "node:fs";
+
+export async function lintHtml(resourceName: string, contentStream: ReadStream): Promise<TranspileResult> {
+	const taskLintEnd = taskStart("Static lint", resourceName);
```
RandomByte, line 10:
```suggestion
	const taskLintEnd = taskStart("Linting HTML", resourceName);
```

tsconfig.base.json
RandomByte: This config is not used in the type checker, but for the UI5 linter sources.

You need to configure this in `src/detectors/typeChecker/index.ts`

RandomByte (changes requested):

src/linter/html/HtmlReporter.ts
d3xter666: Ah, thanks!

src/linter/html/HtmlReporter.ts
d3xter666: Thanks!

src/linter/html/linter.ts
d3xter666: Done

src/detectors/transpilers/html/parser.ts
d3xter666: Yep, thanks! It left as I was uncertain where I could get the internal TextNodes. obviously, they are available, after the close tag

tsconfig.base.json
d3xter666: Thanks!

src/detectors/typeChecker/index.ts
d3xter666: Done

test/fixtures/linter/rules/CSPCompliance/NoInlineJS.html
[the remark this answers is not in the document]
d3xter666: <meta charset="UTF-8"> is indeed invalid, but 
<meta charset="UTF-8**/**> is a valid one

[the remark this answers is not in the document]
matz3: If this works, shouldn't we then see some findings in the snapshot result? (e.g. usage of global "sap.ui.model.json.JSONModel")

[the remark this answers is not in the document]
d3xter666: > BTW: https://facebook.github.io/jsx/#sec-jsx suggests that JSX is really XML-oriented, not HTML5. The syntax does not ~foresee~ allow HTML5 void tags like `<meta>`, `<br>` or ``.

Please check my [comment above](https://github.com/SAP/ui5-linter/pull/48#discussion_r1542879246)
It actually is able to do some of the checks, especially the global ones. As far as I get it, it misses the transpilation part that we do prior analyzing files and that's why it's unable to detect deprecated dependencies

src/linter/html/linter.ts
d3xter666: We have agreed on the following:
ruleId "ui5-linter-csp-unsafe-inline-script"
severity: warning
message text: "Use of unsafe inline script"
message details: "{@link topic:fe1a6dba940e479fb7c3bc753f92b28c Content Security Policy}" (links to https://ui5.sap.com/#/topic/fe1a6dba940e479fb7c3bc753f92b28c)

test/fixtures/linter/rules/CSPCompliance/NoInlineJS.html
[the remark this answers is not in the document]
matz3: So why are no findings is reported for this test fixture? Can we enhance it to cover some of the findings that are working?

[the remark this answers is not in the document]
d3xter666: Good point! I suppose it's because of the transpilation phase that's missing here. The access to global variables is wrapped in `sap.ui.controller`. I wonder why `sap.ui.controller` is not reported. Do we make exceptions for certain cases?

However, if you put the `sap.ui.model.json.JSONModel()` directly into the script tag, it's reported correctly.
I'll debug and will create a tests for those cases.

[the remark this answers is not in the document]
d3xter666: I think I got it how it works. The HTML tags are treated as JSX code and respectively converted into NodeObjects. Their contents are treated as text, but depending on how this text is split into chunks, **there's a chance** (but not necessarily) this text also to be parsed and as it is in our case to report issues.
However, this seems very fragile and does not produce constant results.

Given that sometimes it would produce an expected report and sometimes not, for similar inputs I'd suggest disabling this feature and handling that case in the follow-up backlog.

Any objections?

src/detectors/typeChecker/index.ts
```diff
@@ -116,6 +117,10 @@ export class TsProjectDetector extends ProjectBasedDetector {
 					resourcePath = resourcePath.replace(/\.json$/, ".js");
 					const resourceContent = await resource.getString();
 					({source, messages} = await lintManifest(resourcePath, resourceContent));
+				} else if (resourcePath.endsWith(".html")) {
+					resourcePath = resourcePath.replace(/\.html$/, ".jsx");
+					source = await resource.getString();
```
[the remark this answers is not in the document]
d3xter666: https://github.com/SAP/ui5-linter/pull/48#discussion_r1551412367

src/detectors/typeChecker/index.ts
```diff
@@ -116,6 +117,13 @@ export class TsProjectDetector extends ProjectBasedDetector {
 					resourcePath = resourcePath.replace(/\.json$/, ".js");
 					const resourceContent = await resource.getString();
 					({source, messages} = await lintManifest(resourcePath, resourceContent));
+				} else if (resourcePath.endsWith(".html")) {
+					resourcePath = resourcePath.replace(/\.html$/, ".jsx");
```
matz3, line 121: Why is it `jsx` here...

src/detectors/typeChecker/index.ts
```diff
@@ -290,6 +307,17 @@ export class TsFileDetector extends FileBasedDetector {
...
 				internalfilePath = internalfilePath.replace(/\.json$/, ".js");
 				transformationResult = await lintManifest(filePath.replace(/\.json$/, ".js"), fileContent);
+			} else if (filePath.endsWith(".html")) {
+				// TODO: Enable when implement script extraction and parse
+				// Details: TS treats HTML as JSX, but parsing results are not consistent. https://github.com/SAP/ui5-linter/pull/48#discussion_r1551412367
+				// const fileContent = ts.sys.readFile(filePath);
+				// if (!fileContent) {
+				// 	throw new Error(`Failed to read file ${filePath}`);
+				// }
+				internalfilePath = internalfilePath.replace(/\.html$/, ".js");
```
matz3, line 317: ... but `js` here?

src/linter/html/linter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import type {TranspileResult} from "../../detectors/transpilers/AbstractTranspiler.js";
+import type {ReadStream} from "node:fs";
+
+export async function lintHtml(resourceName: string, contentStream: ReadStream): Promise<TranspileResult> {
+	const taskLintEnd = taskStart("Linting HTML", resourceName);
+	const report = new HtmlReporter(resourceName);
+
+	const scriptTags = await extractScriptTags(contentStream);
+	const jsScriptTags = scriptTags.filter((tag) => tag.attributes.every((attr) => {
```
matz3, line 14: Why about doing this filtering already during parsing? Wouldn't this be more efficient?

d3xter666: Done

src/detectors/typeChecker/index.ts
d3xter666: Ah, nice catch! Thanks!
Fixed it

From current implementation's perspective, it doesn't matter. The point is that this file should be either JS or JSX in order to go through the TS parser.

src/detectors/typeChecker/index.ts
d3xter666: Done

test/fixtures/linter/rules/CSPCompliance/NoInlineJS_negative.html
```diff
@@ -0,0 +1,24 @@
...
+			</App>
+		</mvc:View>
+	</script>
+
+	<script type="" src="./path/to/js.js"></script>
+
+	<script src="./another/path/to/js.js"></script>
+</body>
+
+</html>
```
flovogt, line 24: missing empty last line

test/fixtures/linter/rules/CSPCompliance/NoInlineJS.html
```diff
@@ -0,0 +1,40 @@
...
+			},
+			doSomething: function () {
+				alert("Hello World!");
+			}
+		});
+		sap.ui.xmlview({ viewContent: jQuery('#myXml').html() }).placeAt("content");
+	</script>
+</body>
+
+</html>
```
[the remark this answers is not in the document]
d3xter666: Done

test/fixtures/linter/rules/CSPCompliance/NoInlineJS_negative.html
d3xter666: Done

RandomByte (dismissed):
LGTM

src/detectors/transpilers/html/parser.ts
```diff
@@ -0,0 +1,73 @@
...
+			tag.value === "script") {
+			const isJSScriptTag = tag.attributes.every((attr) => {
+				// The "type" attribute of the script tag should be
+				// 1. not set (default),
+				// 2. an empty string,
+				// 3. or a JavaScript MIME type (text/javascript)
+				// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type#attribute_is_not_set_default_an_empty_string_or_a_javascript_mime_type
+				return attr.name.value !== "type" ||
+					(attr.name.value === "type" &&
+					(attr.value.value === "" || attr.value.value === "text/javascript"));
```
codeworrior, line 63: In the scope of SAPUI5, I find ~180 occurrences of the legacy MIME type `application/javascript`. Maybe we should include that one as well? The other legacy types, I didn't find.

According to https://mimesniff.spec.whatwg.org/#javascript-mime-type-essence-match, the MIME types are case insensitive. I don't expect the SAX parser to normalize attribute values, so the check maybe should be case insensitive.

Last but not least: we most likely won't process scripts of type `module` any time soon, but the check for inline scripts could complain about them already (if they're inline)

d3xter666: Thanks, @codeworrior !

I have added the case insensitive checks and also the `application/javascript` type.

Regarding the `module` type, currently it's being ignored as we check only for missing `type` property or against a list of hardcoded, case insensitive (already, thanks!) types.

I have added type="module" to the test cases

src/detectors/transpilers/html/parser.ts
```diff
@@ -60,7 +60,10 @@ export async function extractJSScriptTags(contentStream: ReadStream) {
 				// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type#attribute_is_not_set_default_an_empty_string_or_a_javascript_mime_type
 				return attr.name.value !== "type" ||
 					(attr.name.value === "type" &&
-					(attr.value.value === "" || attr.value.value === "text/javascript"));
+					["",
+						"text/javascript",
+						"application/javascript", /* legacy */
```
RandomByte, line 65: Please also add a test for `application/javascript`

d3xter666: Done

RandomByte (approved):

src/detectors/transpilers/html/parser.ts
matz3: @codeworrior's comment wasn't fully addressed, right? I also agree that the check should complain about type="module", as it is also relevant for CSP unsafe-inline.

In addition, a <script> tag with a "src" attribute and a comment (or even code) as content is falsely reported as finding. Browsers (tested with Chrome only) ignore the script tag content in case a "src" attribute is provided (even without a value).
```html
<script src="foo.js"> // should not be reported as it is not a CSP violation
</script>
```
```html
<script src> // should also not be reported as it is not a CSP violation
console.log("this code won't run");
</script>
```

PR #57, opened 2024-03-29
feat: Enable new ui5lint version in console

d3xter666:
This PR enables notifications in the console when there is a new version of the ui5linter.


d3xter666:
The build is failing because of missing coverage for the newly introduced code of the notifier: https://github.com/SAP/ui5-linter/pull/57/files#diff-fa8d4e24d8399e8350f1c8bad05df53e8032ea995835bf911507015e2db61cddR25

However, it's clear that this is untestable code in terms of unit testing. In `ui5-tooling/cli` this is covered by the integration tests which we're unable to do here (for now), because of the transpilation step.

Adding `istanbul ignore *` solves the issue, but for higher Node versions, for `18.14.2`, that is used for our testing environment it's not taken into account.

I'm wondering what is the possible solution here (apart from lowering the coverage thresholds for that test)?

.nycrc
```diff
@@ -2,6 +2,9 @@
 	"extends": "@istanbuljs/nyc-config-typescript",
 	"reporter": ["lcov", "text", "text-summary"],
 	"include": ["src/**"],
+	"exclude": [
+		"src/cli.ts"
```
d3xter666, line 6: Please check my coment here: https://github.com/SAP/ui5-linter/pull/57#issuecomment-2027064528

I'd appreciate any feedback on this one.

matz3:
The PR title is a little confusing to me. This does not "enable" a version in the console. It shows a hint in case a new version exists. The user still has to install it.

RandomByte (dismissed):
I'm fine with ignoring the coverage decrease for now until we have integration tests.

When rebasing, be aware that we renamed npm-shrinkwrap.json to package-lock.json on main in order to debug a dependabot issue. Also see: https://github.com/SAP/ui5-linter/pull/67

src/cli/middlewares/base.ts
```diff
@@ -1,4 +1,6 @@
 import {initLogger} from "./logger.js";
+import updateNotifier from "./updateNotifier.js";
```
flovogt, line 2: its `ts` not `js` isn't?

d3xter666: Actually, it's JS. We're importing the transpiled modules. Please, check the rest of the imports. It's always `.js`

PR #70, opened 2024-04-10
fix: Respect src attribute for script tags and include module type for checks

d3xter666:
This change addresses the following comments: https://github.com/SAP/ui5-linter/pull/48#discussion_r1556875268


src/linter/html/linter.ts
```diff
@@ -13,8 +13,12 @@ export async function lintHtml(resourceName: string, contentStream: ReadStream):
 
 	jsScriptTags.forEach((tag) => {
 		const scriptContent = tag.textNodes?.map((tNode) => tNode.value).join("").trim();
```
matz3, line 15: The CSP check in browsers seems to behave differently, see:
```html
<!DOCTYPE HTML>
<html>
	<head>
		<meta charset="utf-8">
		<meta http-equiv="Content-Security-Policy" content="default-src 'self'">
		<!-- Refused to execute inline script because it violates the following Content Security Policy directive -->
		<script> </script>
		<!-- No error (empty content) -->
		<script></script>
	</head>
	<body>
	</body>
</html>
```

d3xter666: So, this means, that even whitespaces are treated as non-secure, right?

matz3: A non-empty script tag without a "src" attribute is evaluated, regardless of it's content. And then if there's a CSP policy like above, this will lead to an error.

Note that I didn't read any spec but just tested the behavior in Chrome.

flovogt (commented):
LGTM. @RandomByte could you have also another look?

src/linter/html/linter.ts
```diff
@@ -12,9 +12,13 @@ export async function lintHtml(resourceName: string, contentStream: ReadStream):
...
 	jsScriptTags.forEach((tag) => {
-		const scriptContent = tag.textNodes?.map((tNode) => tNode.value).join("").trim();
+		const scriptContent = tag.textNodes?.map((tNode) => tNode.value).join("");
+		// Tags with src attribute do not parse and run inline code
+		const hasSrc = tag.attributes.some((attr) => {
+			return attr.name.value.toLowerCase() === "src";
+		});
 
-		if (scriptContent) {
+		if (!hasSrc && scriptContent) {
```
RandomByte, line 21: Do you think it could be enough to check for the existence of any text node now?
If yes, we could remove the map/join calls above.

```suggestion
		if (tag.TextNodes?.length > 0 && scriptContent) {
```

RandomByte (changes requested):

src/linter/html/linter.ts
d3xter666: Hmm, thanks!

d3xter666: Done with https://github.com/SAP/ui5-linter/pull/70/commits/4265bfb280f1ddf4ada8951727b41e6b1179fe84

RandomByte (approved):

PR #73, opened 2024-04-16
feat: Component best practices- async flags check

d3xter666:
JIRA: CPOUI5FOUNDATION-792


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -449,6 +460,164 @@ export default class SourceFileLinter {
...
+			}
+		}
+
+		if (!hasAsyncInterface) {
+			if (rootViewAsyncFlag !== true || routingAsyncFlag !== true) {
+				this.#reporter.addMessage({
+					node: classDesc,
+					severity: LintMessageSeverity.Error,
+					ruleId: "ui5-linter-no-sync-loading",
+					message: "Use of sync loading for Component's views",
```
matz3, line 554: It would be very helpful to get a better message in such case. I have tested this on a project that does not use routing (openui5-sample-app) and was confused that there was an error.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -449,6 +460,164 @@ export default class SourceFileLinter {
...
+				instanceOfPropsRecord(manifestSapui5Section?.routing?.value) &&
+				instanceOfPropsRecord(manifestSapui5Section?.routing?.value.config?.value) &&
+				typeof manifestSapui5Section?.routing?.value.config?.value.async?.value === "boolean") {
+				routingAsyncFlag = manifestSapui5Section?.routing?.value.config?.value.async?.value;
+				routingAsyncFlagNode = manifestSapui5Section?.routing?.value.config?.value.async?.node;
+			}
+		}
+
+		if (!hasAsyncInterface) {
+			if (rootViewAsyncFlag !== true || routingAsyncFlag !== true) {
```
matz3, line 549: If there is no rootView or routing defined, the values must not be checked. Both sections are optional. This currently causes false-positives.

d3xter666: Thanks!
Fixed it

test/fixtures/linter/rules/BestPractices/Negative_1/Component.js
```diff
@@ -0,0 +1,10 @@
+sap.ui.define(["sap/ui/core/UIComponent"], function (UIComponent) {
+	"use strict";
+
+	return UIComponent.extend("com.ui5.troublesome.app.Component", {
+		interfaces: ["sap.ui.core.IAsyncContentCreation"],
```
matz3, line 5: As per https://ui5.sap.com/1.120.12/#/topic/676b636446c94eada183b1218a824717 and https://ui5.sap.com/1.120.12/#/topic/27ce0e4987cd426f8fa3e60836316428 the `interfaces` are defined within the `metadata`.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -449,6 +460,168 @@ export default class SourceFileLinter {
 		}
 	}
 
+	analyzeComponentJson(node: ts.ExpressionWithTypeArguments) {
+		if (node.expression.getText() !== "UIComponent") {
```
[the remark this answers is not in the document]
d3xter666: I'm not sure whether we can handle such situations.
The reasoning is that we do not have a mechanism to analyse the content of the dependencies so far. 
Further more, the dependency could be an external lib/app and in such cases we cannot even access the code/type definitions for it.

I saw the `sap/fe/core/AppComponent`. But as long as it has this interface everything is clear there and we shouldn't report anything. Its children would inherit that and our check would not analyze them as they do not inherit from `UIComponent`
So far the hardcoded option could be the only solution and the benefit from it would be to warn developers when they have `async` flags in their manifests, so they can clean up.
The real issue would have been if the `sap/fe/core/AppComponent` didn't have that interface, but other components were inheriting from it. Then, maybe the hardcoded solution would be the only option so far.

Any feedback on how to workaround this limitation would be highly appreciated!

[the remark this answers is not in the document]
d3xter666: > the AppComponent class is special in a way as it implements the "IAsyncContentCreation" interface, but that interface is not reflected in sap.fe.core.d.ts (to me it is currently unknown why this is the case). 

I suppose that it's because of the UI5 version. Maybe, they have impolemented that interface after 1.120.x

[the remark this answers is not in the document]
matz3: > I suppose that it's because of the UI5 version. Maybe, they have impolemented that interface after 1.120.x

The interface is defined there already for a while, so I suspect it's an issue with generating the d.ts.

> The real issue would have been if the sap/fe/core/AppComponent didn't have that interface, but other components were inheriting from it. Then, maybe the hardcoded solution would be the only option so far.

Then let's use [APF](https://ui5.sap.com/#/topic/0109e67712bc4966bfa8922909e92a87) as another example of a base component class. It does not define the interface and ideally we should detect component best practices for that use case.
I hope you get my point. We should check the inheritance chain to see whether a `Component.js` module inherits from `UIComponent`.

test/fixtures/linter/rules/BestPractices/Negative_1/Component.js
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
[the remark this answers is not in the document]
d3xter666: Now the soluton uses type definitions and is able to go down to the Base class. As we need to detect only `UIComponent`, the soltion is limited up to it.

test/lib/linter/rules/snapshots/CSPCompliance.ts.md
```diff
@@ -79,3 +79,79 @@ Generated by [AVA](https://avajs.dev).
 > Snapshot 1
 
     []
+
+## General: CSPCompliance/NoInlineJS.html
```
matz3, line 83: Why is there new snapshot data for unrelated checks?

test/lib/linter/rules/snapshots/BestPractices.ts.md
```diff
@@ -0,0 +1,271 @@
...
+## General: Negative_5/Component.js
+
+> Snapshot 1
+
+    [
+      {
+        coverageInfo: [],
+        errorCount: 0,
+        fatalErrorCount: 0,
+        filePath: 'Negative_5/Component.js',
```
matz3, line 76: Why are there two files with the same path listed in the results?

test/lib/linter/rules/snapshots/BestPractices.ts.md
```diff
@@ -0,0 +1,271 @@
...
+        coverageInfo: [],
+        errorCount: 0,
+        fatalErrorCount: 0,
+        filePath: 'Positive_2/Component.js',
+        messages: [
+          {
+            column: 9,
+            fatal: undefined,
+            line: 4,
+            message: 'Remove the async flag for "sap.ui5/rootView" from the manifest',
```
matz3, line 153: What about a more positive message here, that also explains **why** the flag can/should be removed?
Without seeing the reference to `IAsyncContentCreation` I doubt that users would understand why they should remove an async flag, as all of our other documentation and best practices tell developers to make use of async flags.

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,271 @@
...
+			classDefinition?.heritageClauses?.flatMap((parentClasses: ts.HeritageClause) => {
+				return parentClasses.types.flatMap((parentClass) => {
+					const parentClassType = checker.getTypeAtLocation(parentClass);
+
+					return parentClassType.symbol?.declarations?.flatMap((declaration) => {
+						let result = {...returnTypeTemplate} as AsyncInterfaceFindType;
+						// Continue down the heritage chain to search for
+						// the async interface or manifest flags
+						if (ts.isClassDeclaration(declaration) &&
+							declaration?.name?.getText() !== "UIComponent") {
```
matz3, line 100: What about a `mycomp.base.UIComponent` class that itself inherits from another class? Isn't this check a bit too vague, as it only checks for the class name, but not in which namespace this class exists?

src/linter/ui5Types/SourceFileLinter.ts
matz3: My point here was not only the false-positive, but also the fact that it mentions "Root View **and Routing** are not configured to load targets asynchronously", even when my `manifest.json` does not use any routing features.

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,271 @@
...
+	// undefined has ambiguous meaning in that context.
+	// It could mean either implicit "true" or "false".
+	// To distinguish whether it's been set from manifest's config
+	// or not set at all, we'll use null.
+	let rootViewAsyncFlag: boolean | undefined | null = null;
+	let routingAsyncFlag: boolean | undefined | null = null;
+
+	if (componentManifest && ts.isPropertyAssignment(componentManifest)) {
+		// The manifest is an external manifest.json file
+		if (componentManifest.initializer.getText() === "\"json\"") {
```
matz3, line 148: While this is technically correct, we decided for a different handling here: `If in the Component.js no object and no "manifest: 'json'" is declared, "manifest.json" should be still analyzed`. This is because when using manifest-first, the `manifest.json` file is loaded first and is still used, even when the Component class doesn't reference it. This is obviously not correct and a reference should be maintained in the metadata. However, such cases may appear and the best practices should also be checked for them.

To improve this situation, we should add a general rule that checks for this consistency. I added it to my list.

matz3 (commented):
I don't see test cases for the other common base component classes in UI5:
- sap/ui/generic/app/AppComponent
- sap/suite/ui/generic/template/lib/AppComponent
- sap/fe/core/AppComponent

d3xter666:
> src/linter/ui5Types/BestPractices.ts

These are all extending the `sap/ui/core/UIComponent`, so the general solution would handle them with the help of UI5 Types. Further more, if some of those classes starts implementing the async interface, that would also be matched

src/linter/ui5Types/BestPractices.ts
d3xter666: Done with https://github.com/SAP/ui5-linter/pull/73/commits/423d76671e8632ffdd824e50d2233678257bf9be

test/lib/linter/rules/snapshots/BestPractices.ts.md
d3xter666: Done

test/lib/linter/rules/snapshots/CSPCompliance.ts.md
d3xter666: Fixed it

test/lib/linter/rules/snapshots/BestPractices.ts.md
d3xter666: Fixed it. It was because the tests were using testnames for that

src/linter/ui5Types/BestPractices.ts
d3xter666: Thanks! Now the manifest gets always analyzed and when there's no definition for a manifest in the Component's metadata (either inline or external), there would be a warning for that

src/linter/ui5Types/SourceFileLinter.ts
matz3: This is still open. The message might suggest to a user that they have to specify "sap.ui5/routing/config" in order to solve this issue. But if routing is not used, the routing configuration does not need to be provided at all.

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+
+	if (!hasAsyncInterface) {
+		// undefined has ambiguous meaning in that context.
+		// It could mean either implicit "true" or "false".
+		// To distinguish whether it's been set from manifest's config
+		// or not set at all, we'll use null.
+		if (rootViewAsyncFlag === false || rootViewAsyncFlag === undefined ||
+			routingAsyncFlag === false || routingAsyncFlag === undefined ||
+			(hasAsyncInterface === null && rootViewAsyncFlag === null && routingAsyncFlag === null)) {
+			reporter.addMessage({
```
matz3, line 313: Currently, this message is also produced when the interface is not defined and no sap.ui5/rootView and no sap.ui5/routing sections exist. In such a case there should not be any message, as no sync behavior is triggered.

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+
+function reportResults(
+	analysisResult: AsyncInterfaceFindType,
+	reporter: SourceFileReporter,
+	classDesc: ts.ClassDeclaration
+) {
+	const {hasAsyncInterface, routingAsyncFlag, rootViewAsyncFlag, hasManifestDefinition} = analysisResult;
+
+	if (!hasManifestDefinition) {
+		reporter.addMessage({
```
matz3, line 295: There's a false positive finding when the manifest is declared this way:
```js
metadata: {
  "manifest": "json"
}
```
and no finding (expected) when it's defined like this
```js
metadata: {
  manifest: "json"
}
```

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+		classDefinition?.heritageClauses?.flatMap((parentClasses: ts.HeritageClause) => {
+			return parentClasses.types.flatMap((parentClass) => {
+				const parentClassType = checker.getTypeAtLocation(parentClass);
+
+				return parentClassType.symbol?.declarations?.flatMap((declaration) => {
+					let result = {...returnTypeTemplate} as AsyncInterfaceFindType;
+					// Continue down the heritage chain to search for
+					// the async interface or manifest flags
+					if (ts.isClassDeclaration(declaration) &&
+						(!uiComponentImportVar || declaration?.name?.getText() !== uiComponentImportVar)) {
```
RandomByte, line 118: Some comments would be helpful for future readers. At least I'm a bit lost here.

`uiComponentImportVar` holds the name of the import variable for `sap/ui/core/UIComponent´ in the Component.js file that is being analyzed? Why is this relevant when analyzing the super classes?

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+			});
+		}) ?? [];
+
+	return [...heritageAnalysis, curClassAnalysis].reduce((acc, curAnalysis) => {
+		return mergeResults(acc ?? {...returnTypeTemplate}, curAnalysis ?? {...returnTypeTemplate});
+	});
+}
+
+function doAsyncInterfaceChecks(importDeclaration: ts.Node): boolean | undefined {
+	while (!importDeclaration || importDeclaration.kind !== ts.SyntaxKind.SourceFile) {
```
RandomByte, line 140: Wouldn't `!importDeclaration` lead to an exception once a node has an undefined parent (unable to access parent on it)? Am I missing something? 😅 

Did you mean something like:
```suggestion
	while (importDeclaration && importDeclaration.kind !== ts.SyntaxKind.SourceFile) {
```

RandomByte (changes requested):

src/linter/ui5Types/BestPractices.ts
d3xter666: Thanks! Fixed with https://github.com/SAP/ui5-linter/pull/73/commits/af1d2f9d21cb8a083405848b3a5f6bf27f0866b4

src/linter/ui5Types/BestPractices.ts
d3xter666: Fixed with https://github.com/SAP/ui5-linter/pull/73/commits/29fa1dbbcf4cb0f9ea0c837cb39a794c69560338 https://github.com/SAP/ui5-linter/pull/73/commits/a7a307e6ddc459657676945b71324771de4292b7 https://github.com/SAP/ui5-linter/pull/73/commits/79cdbe79bb49518d42a412723653ff2333ca3e21 https://github.com/SAP/ui5-linter/pull/73/commits/89f11f8c964dff363f66e68f4bddd34bd77cd52f

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+	const moduleImports = parent.imports as ts.Node[];
+	const uiComponentImportVar = moduleImports.reduce((varName: string, importClause: ts.Node) => {
+		if (ts.isIdentifier(importClause) && importClause.text === "sap/ui/core/UIComponent" &&
+			ts.isImportDeclaration(importClause.parent)) {
+			varName = importClause.parent?.importClause?.name?.getText() ?? "";
+		}
+		return varName;
+	}, "");
+
+	if (classDesc && ts.isClassDeclaration(classDesc)) {
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/ui5Types/BestPractices.ts
```diff
@@ -0,0 +1,345 @@
...
+	};
+}
+
+function findAsyncInterface({classDefinition, manifestContent, checker, uiComponentImportVar}: {
+	classDefinition: ts.ClassDeclaration;
+	manifestContent: string | undefined;
+	checker: ts.TypeChecker;
+	uiComponentImportVar: string;
+}): AsyncInterfaceFindType | undefined {
+	if (!ts.isClassDeclaration(classDefinition)) {
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/linter/ui5Types/BestPractices.ts
d3xter666: Done

src/linter/ui5Types/BestPractices.ts
d3xter666: Thanks! It's not needed anymore for the purpose it served. The main reason behind it was that it was used to not go deeper than UIComponent.

src/linter/ui5Types/TypeLinter.ts
```diff
@@ -103,12 +103,20 @@ export default class TypeChecker {
 				if (!sourceMap) {
 					log.verbose(`Failed to get source map for ${sourceFile.fileName}`);
 				}
+				let manifestContent;
+				if (sourceFile.fileName.endsWith("/Component.js")) {
```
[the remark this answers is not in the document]
d3xter666: Thanks! I was thinking the same, but AppComponents, for example, got me confused 😄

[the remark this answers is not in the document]
matz3: Those are fine as they are not an entry-point for an actual instance of a component.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks! Fixed it with https://github.com/SAP/ui5-linter/pull/73/commits/cacf9007fbdae7bef2423685eae7c8feeae952e0

src/linter/ui5Types/BestPractices.ts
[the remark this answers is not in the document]
RandomByte: I think the rename of the file itself did not happen or did not work

src/linter/ui5Types/bestPractices.ts
```diff
@@ -0,0 +1,358 @@
...
+			if (AsyncInterfaceStatus.parentPropNotSet !== routingAsyncFlag) {
+				componentsToAddress.push("Routing");
+			}
+
+			reporter.addMessage({
+				node: classDesc,
+				severity: LintMessageSeverity.Error,
+				ruleId: "ui5-linter-no-sync-loading",
+				message: `${componentsToAddress.join(" and ")} ${componentsToAddress.length > 1 ? "are" : "is"} ` +
+				"not configured to load targets asynchronously",
```
matz3, line 330: Please review the text. `Root View is not configured to load targets asynchronously` is confusing to me. For the root view there are no "targets", it's just the view. To me it seems this wording only applies to the routing targets.

src/linter/ui5Types/bestPractices.ts
```diff
@@ -0,0 +1,358 @@
...
+				messageDetails: "{@link topic:676b636446c94eada183b1218a824717 Use Asynchronous Loading}. " +
+				"Implement sap.ui.core.IAsyncContentCreation interface in Component.js or set async flags for " +
+				"\"sap.ui5/routing/config\" and \"sap.ui5/rootView\" in the manifest.json",
```
matz3, line 333: Details still always mention both, which might be confusing.

src/linter/ui5Types/bestPractices.ts
```diff
@@ -0,0 +1,358 @@
...
+			});
+		}
+	} else {
+		if (rootViewAsyncFlag === AsyncInterfaceStatus.true) {
+			reporter.addMessage({
+				node: classDesc,
```
matz3, line 339: Wouldn't it be a lot more intuitive for developers to see those messages within the `manifest.json` at the exact line where the flag should be removed? Right now all messages are reported in the Component.js which can be confusing. I think that's fine for the other messages, as the preferred solution is to add the interface in Component.js, but in this case a change should only be made in manifest.json

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -67,6 +70,13 @@ export default class SourceFileLinter {
 				node as (ts.PropertyAccessExpression | ts.ElementAccessExpression)); // Check for deprecation
 		} else if (node.kind === ts.SyntaxKind.ImportDeclaration) {
 			this.analyzeImportDeclaration(node as ts.ImportDeclaration); // Check for deprecation
+		} else if (node.kind === ts.SyntaxKind.ExpressionWithTypeArguments) {
+			analyzeComponentJson(
```
matz3, line 74: Just some idea on improving runtime performance:
Currently, this function is called for every class that extends from another class. Just to then go up the parent chain to find out that the file we're looking at is not a component. Wouldn't it be better to do the check whether we are looking at a Component at the time we are visiting the `SourceFile` node and then remember it for later? This would prevent the unnecessary overhead of always going up to the SourceFile node for every class.

src/linter/ui5Types/BestPractices.ts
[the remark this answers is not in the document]
d3xter666: Now it's renamed. Thanks!

src/linter/linter.ts
```diff
@@ -89,7 +89,7 @@ export async function lintFile({
 }: LinterOptions): Promise<LintResult[]> {
 	const reader = createReader({
 		fsBasePath: rootDir,
-		virBasePath: "/",
+		virBasePath: namespace ? `/resources/${namespace}/` : "/",
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/ui5Types/bestPractices.ts
d3xter666: Done with https://github.com/SAP/ui5-linter/pull/73/commits/b3c1bb53805268957f238249fbb97a74b74f25da

src/linter/ui5Types/bestPractices.ts
d3xter666: Refactored with https://github.com/SAP/ui5-linter/pull/73/commits/49d4e33e2377f65700613f2e0dbfba68f332d519

src/linter/ui5Types/bestPractices.ts
d3xter666: Refactored with https://github.com/SAP/ui5-linter/pull/73/commits/49d4e33e2377f65700613f2e0dbfba68f332d519

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done with https://github.com/SAP/ui5-linter/pull/73/commits/34f6698160a145eb4647905e4c489c1ac212a0c1

src/linter/ui5Types/bestPractices.ts
```diff
@@ -314,29 +323,48 @@ function reportResults(
...
 			if (AsyncInterfaceStatus.parentPropNotSet !== rootViewAsyncFlag) {
-				componentsToAddress.push("Root View");
-			}
-			if (AsyncInterfaceStatus.parentPropNotSet !== routingAsyncFlag) {
-				componentsToAddress.push("Routing");
+				// sap.ui5/rootView is not set at all, so skip it in the message
+				message = "Routing is not configured to load its targets asynchronously.";
+				messageDetails = "{@link topic:676b636446c94eada183b1218a824717 Use Asynchronous Loading}. " +
+				"Implement sap.ui.core.IAsyncContentCreation interface in Component.js or set async flag for " +
+				"\"sap.ui5/routing/config\" in the manifest.";
```
matz3, line 336: Are there any tests that cover this behavior? I've tested this with an application and am getting the exact opposite from what I would expect. I have a rootView defined without async flag, and the message complains about a missing flag in the routing. In the Positive_1 fixture both flags are missing, but there's only a message for missing routing flag.

d3xter666: Thanks!
Fixed it with https://github.com/SAP/ui5-linter/pull/73/commits/417a94b0ba2dc8554093c0668c2ca14f7d7ed3d2
Also checked the snapshots and now everything seems to be correct. 

Everything case I imagined is covered now. If you can think of some other, please let me know

matz3: I still see no test case that covers those two messages. Coverage also shows them as not covered. But maybe I'm missing something here. I'm on https://github.com/SAP/ui5-linter/commit/417a94b0ba2dc8554093c0668c2ca14f7d7ed3d2.

d3xter666: Please, have a look at `Positive_2/Component.js`: https://github.com/SAP/ui5-linter/pull/73/files#diff-0ec5155380fd3031aee943bec0ef111257c570eff2d4ceef2cf631ece3c28342R168

matz3: It has different messages. My comment here is about `Routing is not configured [...]` and `Root View is not configured [...]`.

d3xter666: Ah, thanks a lot! I've missed it.

I have added the cases here https://github.com/SAP/ui5-linter/pull/73/commits/604ef2257de05ff4c0fb16d9290d9772ea5218a1

src/linter/ui5Types/bestPractices.ts
```diff
@@ -0,0 +1,381 @@
...
+	reporter: SourceFileReporter;
+	context: LinterContext;
+	classDesc: ts.ClassDeclaration;
+	manifestContent: string | undefined;
+	resourcePath: string;
+}) {
+	const {hasAsyncInterface, routingAsyncFlag, rootViewAsyncFlag, hasManifestDefinition} = analysisResult;
+
+	if (!hasManifestDefinition) {
+		reporter.addMessage({
```
matz3, line 309: This message is also created when there is no `manifest.json` file next to the `Component.js`. If this is not the case, the suggestion of the message does not apply.

d3xter666: Thanks!
In such a case I suppose we need to enforce a fatal error- no definition and no manifest. 
Would UI5 be able to execute the app without a manifest at all?

matz3: A `Component.js` can define a manifest inline, which should not create a message - I think that's covered already.
In addition, there are fields in the `metadata` (not within `manifest`), like `rootView` that can be used and at least my understanding is that although it is not the best practice, it will continue to work with 2.x. So at least for now I wouldn't expect any message in such a case.

d3xter666: Thanks!
Now it's implemented that way.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -67,6 +70,17 @@ export default class SourceFileLinter {
 				node as (ts.PropertyAccessExpression | ts.ElementAccessExpression)); // Check for deprecation
 		} else if (node.kind === ts.SyntaxKind.ImportDeclaration) {
 			this.analyzeImportDeclaration(node as ts.ImportDeclaration); // Check for deprecation
+		} else if (node.kind === ts.SyntaxKind.ExpressionWithTypeArguments &&
+		ts.isSourceFile(this.#sourceFile) &&
+		this.#sourceFile.fileName.endsWith("/Component.js")) {
```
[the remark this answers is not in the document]
RandomByte: Done

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -67,6 +70,17 @@ export default class SourceFileLinter {
 				node as (ts.PropertyAccessExpression | ts.ElementAccessExpression)); // Check for deprecation
 		} else if (node.kind === ts.SyntaxKind.ImportDeclaration) {
 			this.analyzeImportDeclaration(node as ts.ImportDeclaration); // Check for deprecation
+		} else if (node.kind === ts.SyntaxKind.ExpressionWithTypeArguments &&
+		ts.isSourceFile(this.#sourceFile) &&
```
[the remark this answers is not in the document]
RandomByte: Removed

src/linter/ui5Types/SourceFileLinter.ts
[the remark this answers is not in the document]
RandomByte: Done

src/linter/ui5Types/bestPractices.ts
```diff
@@ -0,0 +1,381 @@
...
+	});
+}
+
+function doAsyncInterfaceChecks(importDeclaration: ts.Node): AsyncInterfaceStatus {
+	while (importDeclaration && importDeclaration.kind !== ts.SyntaxKind.SourceFile) {
+		importDeclaration = importDeclaration.parent;
+	}
+
+	// @ts-expect-error imports is part of SourceFileObject
+	const moduleImports = importDeclaration?.imports as ts.Node[];
```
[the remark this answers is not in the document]
RandomByte: Done

src/linter/linter.ts
```diff
@@ -87,16 +87,19 @@ export async function lintProject({
...
-		virBasePath: "/",
-	});
 	let resolvedFilePaths;
+	let virBasePath = "";
 	if (pathsToLint?.length) {
 		const absoluteFilePaths = resolveFilePaths(rootDir, pathsToLint);
 		resolvedFilePaths = transformFilePathsToVirtualPaths(
 			absoluteFilePaths, rootDir, "/", rootDir);
+		// Extract the (virtual) path from the filename
+		virBasePath = resolvedFilePaths[0].split("/").slice(0, -1).join("/");
```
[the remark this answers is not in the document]
d3xter666: Hmmm, you're right about the deeply nested components 🤔 

But, how to extract the namespace without breaking the signature of the function?
Maybe, reuse the logic from `lintProject()`, but then we need a project, not a single file...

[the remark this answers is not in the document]
RandomByte: I reverted the commit and all tests are still green 😅

RandomByte:
I'm fine with this PR and once https://github.com/SAP/ui5-linter/pull/73#discussion_r1608011338 is clarified we can merge it.

As a follow-up I would like us to extract the texts similar to https://github.com/SAP/ui5-linter/pull/112/files#diff-9d3ca30e9cd9e1355253f085c67fd2abafe4dfff649602cacf8b3ddef81db990 and hand them over to UA for review.

src/linter/linter.ts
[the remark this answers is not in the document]
d3xter666: Please take a look here: [src/linter/linter.ts](https://github.com/SAP/ui5-linter/pull/73/files#diff-b5f6735c1e11d603b4261047a776b51e3fe83892f36e19a6a0506f59ab152032L92)

This is the original change compared to the main branch.
The change is working pretty well, except that it cannot handle `/test-resources/` and that's why I needed to find a better way of doing it.
Please take a look at the conversation we had with @flovogt on that matter here: https://github.com/SAP/ui5-linter/pull/73/files#r1601613818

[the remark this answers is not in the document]
RandomByte: Got it, thanks. Since the `lintFile` API is currently only used in our tests, I think we can safely go with the current state and ignore the test-resources case for now 👍

RandomByte (approved):

RandomByte:
GitHub requires an additional review from someone other than me 😅 @flovogt

flovogt (approved):

PR #90, opened 2024-04-23
fix: Don't exit synchronously when there still might be async I/O

codeworrior:
See https://nodejs.org/api/process.html#processexitcode for an explanation of process.exit and process.exitCode and https://nodejs.org/api/process.html#a-note-on-process-io for an explanation when I/O might be sync or async.

Additionally writes a newline after format "json" has been written. This prevents strange behavior in at least zsh when output is not terminated by a newline.

### DIFF
--- a/test/lib/cli/base.integration.ts
+++ b/test/lib/cli/base.integration.ts
@@ -38,14 +38,20 @@ test.serial("ui5lint --format json", async (t) => {
 
 	t.is(consoleLogStub.callCount, 0, "console.log should not be used");
 	t.true(processCwdStub.callCount > 0, "process.cwd was called");
-	t.is(processStdoutWriteStub.callCount, 1, "process.stdout.write is only used once");
-	t.is(processExitStub.callCount, 1, "process.exit got called once");
-	t.is(processExitStub.firstCall.firstArg, 1, "process.exit got called with exit code 1");
+	t.is(processStdoutWriteStub.callCount, 2, "process.stdout.write was called twice");
+	t.is(processExitStub.callCount, 0, "process.exit got never called");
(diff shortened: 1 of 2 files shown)


test/lib/cli/base.integration.ts
```diff
@@ -38,14 +38,20 @@ test.serial("ui5lint --format json", async (t) => {
...
 	t.is(consoleLogStub.callCount, 0, "console.log should not be used");
 	t.true(processCwdStub.callCount > 0, "process.cwd was called");
-	t.is(processStdoutWriteStub.callCount, 1, "process.stdout.write is only used once");
-	t.is(processExitStub.callCount, 1, "process.exit got called once");
-	t.is(processExitStub.firstCall.firstArg, 1, "process.exit got called with exit code 1");
+	t.is(processStdoutWriteStub.callCount, 2, "process.stdout.write was called twice");
+	t.is(processExitStub.callCount, 0, "process.exit got never called");
+	t.is(process.exitCode, 1, "process.exitCode was set to 1");
+	// cleanup: reset exit code in order not to fail the test (it cannot be stubbed with sinon)
+	process.exitCode = 0;
```
codeworrior, line 45: I didn't like to manipulate the "real" exitCode, but sinon could not stub it as the property descriptor is "configurable: false".

You might check with
```
node -p 'Object.getOwnPropertyDescriptor(process, "exitCode")'
```

Manipulating the exitCode somewhat seems risky to me as I don't know whether ava uses that property, too.

Any opinions on that?

RandomByte: Fine with me. If it works now I would expect it to keep working like this for a while 🙂

RandomByte (approved):
LGTM. I guess we should make the same change in the UI5 CLI

PR #104, opened 2024-05-07
feat: Add detection for deprecated dependencies in .library

maxreichmann:
JIRA: CPOUI5FOUNDATION-825

+refactor: Add common `xmlParser`
+refactor: Remove `htmlParser`

The `Lib.init()` call detection is addressed in the following PR: https://github.com/SAP/ui5-linter/pull/197


src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,65 @@
...
+	async lint() {
+		try {
+			const dotLibraryDependencyTags = await this.#parseDotLibrary(this.#contentStream);
+			this.#analyzeDeprecatedLibs(dotLibraryDependencyTags);
+		} catch (err) {
+			const message = err instanceof Error ? err.message : String(err);
+			this.#context.addLintingMessage(this.#resourcePath, {
+				severity: LintMessageSeverity.Error,
+				message,
+				ruleId: "ui5-linter-parsing-error",
```
d3xter666, line 28: Please use the `formatMessage` util
as it's done here: https://github.com/SAP/ui5-linter/pull/197/files#diff-47ae50dc9ac33799e2dd8a28629230419b8dcb6e7c0492f17801a53f6b59d7ceR96

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,65 @@
...
+			const {line, character: column} = lib.openStart;
+
+			if (deprecatedLibraries.includes(libName)) {
+				this.#context.addLintingMessage(this.#resourcePath, {
+					ruleId: "ui5-linter-no-deprecated-api",
+					severity: LintMessageSeverity.Error,
+					fatal: undefined,
+					line: line + 1,
+					column: column + 1,
+					message: `Use of deprecated library '${libName}'`,
```
d3xter666, line 60: Please use the `formatMessage` util
as it's done here: https://github.com/SAP/ui5-linter/pull/197/files#diff-47ae50dc9ac33799e2dd8a28629230419b8dcb6e7c0492f17801a53f6b59d7ceR96

maxreichmann: Resolved by [8f7284a](https://github.com/SAP/ui5-linter/pull/104/commits/8f7284a9fbdd0db544bbe5550c95b02a91e6113f)

src/linter/dotLibrary/DotLibraryLinter.ts
maxreichmann: Resolved by [8f7284a](https://github.com/SAP/ui5-linter/pull/104/commits/8f7284a9fbdd0db544bbe5550c95b02a91e6113f)

maxreichmann: Resolved?

d3xter666 (commented):
LGTM, but we worked in pair on that, so someone else needs to take a look, too

src/linter/dotLibrary/linter.ts
```diff
@@ -0,0 +1,29 @@
+import {LinterParameters} from "../LinterContext.js";
+import DotLibraryLinter from "./DotLibraryLinter.js";
+import {Resource} from "@ui5/fs";
+
+export default async function lintDotLibrary({context}: LinterParameters) {
+	let dotLibraryResources: Resource[];
+	const pathsToLint = context.getPathsToLint();
+	const reader = context.getRootReader();
```
flovogt, line 8: do we need the rootreader here? we could use the `workspace` as we do not expect .library files on the root level, isn't?

matz3:
When limiting the linting to just the `.library`, the linter fails, which might be related to @flovogt's comment about the root reader:
```
ui5lint --file-paths src/sap/f/.library
⚠️  Process Failed With Error

Error Message:
Resource not found: /resources/sap/f/.library

Stack Trace:
Error: Resource not found: /resources/sap/f/.library
    at lib/linter/dotLibrary/linter.js:14:23
```

matz3:
When there is an empty node `<libraryName></libraryName>`, there is an exception reported:
```
openui5/src/sap.f/src/sap/f/.library
0:0 error Fatal error: Cannot read properties of undefined (reading 'value')
```

matz3:
The library name node is looked up everywhere in the XML document, not just within the expected structure.
So putting a `<libraryName>sap.ui.commons</libraryName>` anywhere also leads to an error, which isn't really expected. In this case there is no conflict, as the name is not used anywhere else in the document.

Could you check how much effort it would be to really check for the right place of the dependency name in the document?
Only such a definition should be checked:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<library xmlns="http://www.sap.com/sap.ui.library.xsd" >
  <dependencies>
    <dependency>
      <libraryName>sap.ui.commons</libraryName>
    </dependency>
  </dependencies>
</library>

```
As this is not expected to cause any issues, I'm fine with the current state in case this is a high effort to implement.

d3xter666:
> When limiting the linting to just the `.library`, the linter fails, which might be related to @flovogt's comment about the root reader:
> 
> ```
> ui5lint --file-paths src/sap/f/.library
> ⚠️  Process Failed With Error
> 
> Error Message:
> Resource not found: /resources/sap/f/.library
> 
> Stack Trace:
> Error: Resource not found: /resources/sap/f/.library
>     at lib/linter/dotLibrary/linter.js:14:23
> ```

Fixed with https://github.com/SAP/ui5-linter/pull/104/commits/a876b144b3ffbb5e2504af2d240410b504726088

src/linter/dotLibrary/linter.ts
d3xter666: Thanks!

d3xter666:
> not read properties of undefined (reading 'va

Fixed

d3xter666:
> The library name node is looked up everywhere in the XML document, not just within the expected structure. So putting a `<libraryName>sap.ui.commons</libraryName>` anywhere also leads to an error, which isn't really expected. In this case there is no conflict, as the name is not used anywhere else in the document.
> 
> Could you check how much effort it would be to really check for the right place of the dependency name in the document? Only such a definition should be checked:
> 
> ```
> <?xml version="1.0" encoding="UTF-8" ?>
> <library xmlns="http://www.sap.com/sap.ui.library.xsd" >
>   <dependencies>
>     <dependency>
>       <libraryName>sap.ui.commons</libraryName>
>     </dependency>
>   </dependencies>
> </library>
> ```
> 
> As this is not expected to cause any issues, I'm fine with the current state in case this is a high effort to implement.

We've managed to handle this one with https://github.com/SAP/ui5-linter/pull/104/commits/12bfaec529e562dff2831754e1c05108a1b0f295

src/linter/dotLibrary/linter.ts
```diff
@@ -0,0 +1,28 @@
...
+		dotLibraryResources = [];
+		await Promise.all(pathsToLint.map(async (resourcePath) => {
+			if (!resourcePath.endsWith(".library")) {
+				return;
+			}
+			const resource = await workspace.byPath(resourcePath);
+			if (!resource) {
+				throw new Error(`Resource not found: ${resourcePath}`);
+			}
+			dotLibraryResources.push(resource);
```
matz3, line 18: This is not covered by the tests

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,99 @@
...
+			const dotLibraryDependencyTags = await this.#parseDotLibrary(this.#contentStream);
+			this.#analyzeDeprecatedLibs(dotLibraryDependencyTags);
+		} catch (err) {
+			const message = err instanceof Error ? err.message : String(err);
+			this.#context.addLintingMessage(this.#resourcePath, {
+				severity: LintMessageSeverity.Error,
+				message,
+				ruleId: RULES["ui5-linter-parsing-error"],
+				fatal: true,
+			});
```
matz3, line 31: Not covered by unit tests

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+			if (event === SaxEventType.OpenTag && !tag.selfClosing) {
+				tagsStack.push(tag.value);
+			} else if (event === SaxEventType.CloseTag && !tag.selfClosing) {
+				tagsStack.pop();
+			}
+
+			if (event === SaxEventType.CloseTag &&
+				tag.value === "libraryName") {
+				const path = tagsStack.slice(-1 * libNamePath.length);
```
matz3, line 52: When changing this to
```js
const path = tagsStack;
```

all tests still succeed. In addition, the path is only matched when there are exactly 3 tags in the stack. To me this matches with the expected behavior.

Otherwise (with the slice), the path would still be matched when there are more tags at root level (e.g. `foo > library > dependencies > dependency`).

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	#analyzeDeprecatedLibs(libs: SaxTag[]) {
+		// Check for deprecated libraries
+		libs.forEach((lib) => {
+			const {line, character: column} = lib.openStart;
+
+			// Check for missing textNodes or for empty textNodes withing the
+			// libraryName tag
+			if (!lib.textNodes[0] || lib.textNodes[0].value.trim() === "") {
```
matz3, line 71: I don't expect this to be a common mistake. My comment was rather focused on having a robust check that does not run into unexpected errors / exceptions when the input data is different from the expected cases.

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,99 @@
...
+					line: line + 1,
+					column: column + 1,
+					message: formatMessage(MESSAGES.SHORT__EMPTY_LIBRARY_NAME),
+					messageDetails: formatMessage(MESSAGES.DETAILS__EMPTY_LIBRARY_NAME),
+				});
+
+				return;
+			}
+
+			const libName = lib.textNodes[0].value.trim();
```
matz3, line 85: As UI5 Tooling doesn't seem to trim values when generating a manifest.json from a .library file, I would suggest that we're consistent here and also do not trim the string.

src/linter/dotLibrary/DotLibraryLinter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+			if (deprecatedLibraries.includes(libName)) {
+				this.#context.addLintingMessage(this.#resourcePath, {
+					ruleId: RULES["ui5-linter-no-deprecated-api"],
```
matz3, line 89: That is not the ID that corresponds to the message, right? To me it seems the `addLintingMessage` API should become easier to use. Right now it's error prone to manually format the message and use the correct corresponding ruleId.

matz3 (commented):
LGTM. The mentioned issue have been addressed. Just some minor remarks.

src/linter/dotLibrary/DotLibraryLinter.ts
d3xter666: Thanks! Fixed the ruleId

Maybe, we can discuss how to handle this as I see many ways to do it and it'd always require some level of attention and detail

src/linter/dotLibrary/DotLibraryLinter.ts
d3xter666: Done

src/linter/dotLibrary/DotLibraryLinter.ts
matz3: Yes, it's not completely straight-forward. I'd like to hear your ideas.

src/linter/dotLibrary/DotLibraryLinter.ts
d3xter666: Removed this and provided a check

src/linter/dotLibrary/DotLibraryLinter.ts
d3xter666: Done

src/linter/dotLibrary/DotLibraryLinter.ts
d3xter666: At first sight I was thinking about having some kind of currying, so that context's `addLinteMessage` would come in side the specific linting module preconfigured. However there are cases where we detect multiple things and that might be the challenge here.

matz3 (approved):

src/linter/dotLibrary/DotLibraryLinter.ts
matz3: I've created https://github.com/SAP/ui5-linter/pull/251 with some rough idea I had. But we should probably also think about introducing a base class for the reporters so that we can streamline the signatures.

I also had the idea to only collect the information necessary to create the messages during linting, and only produce the text and details when creating the report. This could reduce the memory usage during linting, which is already very high. The enum would help with that, as it is technically just a number in JS.
But this can easily be done in a follow-up refactoring, as it most probably won't affect the `addMessage` signature.

PR #127, opened 2024-05-27
refactor: Refactoring of detect Component best practices PR

RandomByte:


RandomByte:
Test case `Positive_11` seems to not detect the missing async flags/interface. Need to look into this tomorrow

src/linter/ui5Types/asyncComponentFlags.ts
```diff
@@ -118,38 +117,58 @@ function findAsyncInterface({classDefinition, manifestContent, checker}: {
...
+	}
+	const hasAsyncImport = coreLibImports.some((importDecl) => {
+		const importClause = importDecl.importClause;
+		if (!importClause) {
+			return;
+		}
+		if (!importClause.namedBindings) {
+			// Example: import "sap/ui/core/library"; or import library from "sap/ui/core/library";
+		} else if (ts.isNamedImports(importClause.namedBindings)) {
+			// Example: import { IAsyncContentCreation } from "sap/ui/core/library";
```
d3xter666, line 162: There's also an edge case here- to rename the import 😅 
`import { IAsyncContentCreation: myRenamedAsyncFlag } from ...`

d3xter666:
> Test case `Positive_11` seems to not detect the missing async flags/interface. Need to look into this tomorrow

The issue is that the `Component` does not have `members` (https://github.com/SAP/ui5-linter/pull/127/files#diff-b33fcef6fb0b205372cd4799b7486a23abf3f17b4461bc35e21a0466ad1e9179R89), therefore the check for the manifest.json gets skipped.
The manifest is not propagetd to the parent component analysis and it never got checked: https://github.com/SAP/ui5-linter/pull/127/files#diff-b33fcef6fb0b205372cd4799b7486a23abf3f17b4461bc35e21a0466ad1e9179R106
It should have been working before as a side effect of the propagation to the parent component.

RandomByte:
> > Test case `Positive_11` seems to not detect the missing async flags/interface. Need to look into this tomorrow
> 
> The issue is that the `Component` does not have `members` (https://github.com/SAP/ui5-linter/pull/127/files#diff-b33fcef6fb0b205372cd4799b7486a23abf3f17b4461bc35e21a0466ad1e9179R89), therefore the check for the manifest.json gets skipped. The manifest is not propagetd to the parent component analysis and it never got checked: https://github.com/SAP/ui5-linter/pull/127/files#diff-b33fcef6fb0b205372cd4799b7486a23abf3f17b4461bc35e21a0466ad1e9179R106 It should have been working before as a side effect of the propagation to the parent component.

Great, thanks! Changed the fixture, and even fixed a bug `Unhandled CallExpression expression syntax: SuperKeyword`

d3xter666 (approved):

Issue #103, opened 2024-05-07
Linter sometimes reports the wrong origin type for a deprecated method


codeworrior:
## Expected Behavior

When reporting the use of a deprecated method, the linter should name the correct type that provides the method (_Naming the subclass on which the method is called instead of naming the type that defines the method would be acceptable, too_).

## Current Behavior

In the case of method call chains, the linter might name a totally different type, which is confusing for developers.

## Steps to Reproduce the Issue

Lint the following module

```js
sap.ui.define(["sap/m/Button"], (Button) => {
	const ctrl = new Button();
	ctrl.$().control(document.body); // jQuery.fn.control is deprecated
});
```
The linter reports the following
```
"Call to deprecated function 'control' of class 'Button'"
```
but `Button` neither defines nor inherits method `control`. `control` is a method of jQuery.

## Context

- UI5 linter version: **0.2.2**
- Node.js Version: **v20.11.1**
- npm Version: 10.2.4
- OS/Platform: macOS 14.4.1
- Browser *(if relevant)*: n/a
- Other information regarding your environment *(optional)*: n/a

Issue #94, opened 2024-04-26
Line/column position of deprecations is inaccurate


matz3:
## Expected Behavior

Findings should be reported at the exact place where they are in the code.
More specifically, findings for different functions should not be reported at the exact same line/column position.

### Background

Some tools or IDEs only display one finding per line/column, which currently prevents some findings from being displayed.
One example of this are the tasks in VS Code, which can be used to display problems based on the output of a task, like running `ui5lint`.

### Example

The code
```js
sap.ui.getCore().attachInit(() => {});
```
should create two findings
- `Call to deprecated function 'getCore' (sap.ui.getCore)` pointing to column `8` (`getCore`)
- `Call to deprecated function 'attachInit' of class 'Core'` pointing to column `18` (`attachInit`)

```json
[
  {
    "filePath": "webapp/Test.js",
    "messages": [
      {
        "ruleId": "ui5-linter-no-deprecated-api",
        "severity": 2,
        "line": 1,
        "column": 18,
        "message": "Call to deprecated function 'attachInit' of class 'Core'"
      },
      {
        "ruleId": "ui5-linter-no-deprecated-api",
        "severity": 2,
        "line": 1,
        "column": 8,
        "message": "Call to deprecated function 'getCore' (sap.ui.getCore)"
      }
    ],
    "errorCount": 2,
    "warningCount": 0,
    "fatalErrorCount": 0
  }
]
```

## Current Behavior

Both findings point to column 1

```json
[
  {
    "filePath": "webapp/Test.js",
    "messages": [
      {
        "ruleId": "ui5-linter-no-deprecated-api",
        "severity": 2,
        "line": 1,
        "column": 1,
        "message": "Call to deprecated function 'attachInit' of class 'Core'"
      },
      {
        "ruleId": "ui5-linter-no-deprecated-api",
        "severity": 2,
        "line": 1,
        "column": 1,
        "message": "Call to deprecated function 'getCore' (sap.ui.getCore)"
      }
    ],
    "errorCount": 2,
    "warningCount": 0,
    "fatalErrorCount": 0
  }
]
```

## Steps to Reproduce the Issue

See example above.

## Context

- UI5 linter version: **`0.2.2`**
- Node.js Version: **`20.12.2`**
- npm Version: `10.5.0`
- OS/Platform: `macOS`
- Browser *(if relevant)*: `unknown`
- Other information regarding your environment *(optional)*: `nothing`

## Log Output / Stack Trace

See above.

[...]

PR #131, opened 2024-05-31
fix: Error message and position information for deprecated function calls

RandomByte:
When detecting a deprecated function call, correctly determine the type
of the owner of that function.

Fixes https://github.com/SAP/ui5-linter/issues/103

---

Use the node of the actual call instead of the whole expression to
determine the line/column position for the error object.

Fixes https://github.com/SAP/ui5-linter/issues/94


d3xter666 (approved):

matz3 (approved):
LGTM. It introduces some duplicate findings of global access + deprecated API usage, but I think that's fine and it would be very hard to deduplicate all of them.

matz3:
> LGTM. It introduces some duplicate findings of global access + deprecated API usage, but I think that's fine and it would be very hard to deduplicate all of them.

I also just checked whether the deduplication still makes sense and it currently does, for the `jQuery.sap` usage, which would otherwise be reported as global and deprecated usage, which IMO would not be beneficial to the user.

PR #227, opened 2024-07-30
deps: License- add exception for new deps

d3xter666:
All three packages are licensed under the MIT license, but the package.json file is missing the license field.

PRs to add the license field have already been opened for all three packages.

### DIFF
--- a/.licensee.json
+++ b/.licensee.json
@@ -8,6 +8,9 @@
     "blueOak": "bronze"
   },
   "packages": {
+    "atomically": "2.0.3",
+		"stubborn-fs": "1.2.5",
+		"when-exit": "2.1.3",
     "@sapui5/types": "^1.120.4",
     "callsite": "1.0.0",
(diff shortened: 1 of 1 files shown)


.licensee.json
```diff
@@ -8,6 +8,9 @@
     "blueOak": "bronze"
   },
   "packages": {
+    "atomically": "2.0.3",
+		"stubborn-fs": "1.2.5",
```
flovogt, line 12: tab issue

d3xter666: Thanks!

flovogt (approved):

matz3:
I would rather categorize this as `test`, because it's not relevant for the consumers so no need to have it in the release notes.

flovogt:
> I would rather categorize this as `test`, because it's not relevant for the consumers so no need to have it in the release notes.
Or `ci(licensee): Add exceptions for new deps`

PR #235, opened 2024-08-01
refactor: Simplify SAPUI5 types update process

matz3:


codeworrior (approved):
Successfully tested locally. Results look as expected.

I struggled a bit with the correct syntax of the domain parameter, but this was due to my laziness (not reading the instructions carefully enough).

Definitely a significant improvement.

d3xter666 (approved):

PR #240, opened 2024-08-02
refactor: Enable eslint RegEx rules

d3xter666:
With the introduction of ESLint 9 and the update of `typescript-eslint`, there are new rules that will be applied to the code. Those rules are mainly focused on Regex patterns.

With https://github.com/SAP/ui5-linter/pull/230 those rules have been turned off to keep it consistent with the old behavior.
This change aims to enable those rules.

**Note:** The base for this change is https://github.com/SAP/ui5-linter/pull/230 and not the **main** branch. This way it's clearer what the real changes are.

### DIFF
--- a/src/formatter/lib/resolveLinks.ts
+++ b/src/formatter/lib/resolveLinks.ts
@@ -83,21 +83,25 @@ function _preProcessLinksInTextBlock(sText: string, ui5Url: string, ui5Version: 
 		}
 
 		// If the link has a protocol, do not modify, but open in a new window
-		if (sTarget.includes("://")) {
+		if (/:\/\//.test(sTarget)) {
 			return formatUrlToLink(sTarget, sText);
 		}
 		// topic:xxx Topic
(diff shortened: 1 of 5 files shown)


src/formatter/lib/resolveLinks.ts
```diff
@@ -83,25 +83,21 @@ function _preProcessLinksInTextBlock(sText: string, ui5Url: string, ui5Version:
 		}
 
 		// If the link has a protocol, do not modify, but open in a new window
-		if (/:\/\//.test(sTarget)) {
```
matz3, line 86: Changes in this file should rather be applied within the actual source file, which is located at https://github.com/SAP/openui5/blob/master/lib/jsdoc/transformApiJson.js

For that reason, I wonder whether we should rather disable the rules for this file?

d3xter666: Done

matz3 (dismissed):

matz3 (approved):

PR #251, opened 2024-08-07
refactor(SourceFileReporter): Improve addMessage signature

matz3:


src/linter/linterReporting.ts
```diff
@@ -51,3 +53,16 @@ export function formatMessage(message: string, ...params: string[]) {
 
 	return message;
 }
+
+export enum MESSAGE {
+	NO_DIRECT_DATATYPE_ACCESS,
+}
+
+export const MESSAGE_INFO = {
+	[MESSAGE.NO_DIRECT_DATATYPE_ACCESS]: {
```
d3xter666, line 62: I really like the idea of grouping "dependent" props!

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -529,10 +529,9 @@ export default class SourceFileLinter {
 			if (isDataType) {
 				this.#reporter.addMessage({
 					node: moduleSpecifierNode,
-					severity: LintMessageSeverity.Error,
-					ruleId: RULES["ui5-linter-no-pseudo-modules"],
-					message: formatMessage(MESSAGES.SHORT__NO_DIRECT_DATATYPE_ACCESS, moduleSpecifierNode.text),
-					messageDetails: formatMessage(MESSAGES.DETAILS__NO_DIRECT_DATATYPE_ACCESS, moduleNamespaceName),
+					message: MESSAGE.NO_DIRECT_DATATYPE_ACCESS,
```
d3xter666, line 532: Why changing the signiture as we already have the required props grouped in here.
Maybe, instead we can directly expand?
```suggestion
					...MESSAGE.NO_DIRECT_DATATYPE_ACCESS,
```

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -529,10 +529,9 @@ export default class SourceFileLinter {
...
+					args: [moduleSpecifierNode.text],
```
d3xter666, line 533: What bothers me here is that we will just proxy those to the `formatMessage`.
The needed information is already available here (inside `MESSAGE.NO_DIRECT_DATATYPE_ACCESS`), so can't we somehowe directly apply arguments on top of it via the `formatMessage`?

src/linter/ui5Types/SourceFileLinter.ts
matz3: I think that could be done (but then based on the MESSAGE_INFO). 
owever, it would most probably not allow to reduce memory usage (see my comment https://github.com/SAP/ui5-linter/pull/104#discussion_r1706802866), as objects/strings are already created.
Also, it would again allow providing the properties individually, which is IMHO not necessary and error-prone.

src/linter/ui5Types/SourceFileLinter.ts
matz3: Do you mean to again call `formatMessage` in here?
I'm not in favour of that for two reasons.
1. it would again be a place where things could be out of sync (as at least the message + the enum needs to be provided then)
2. (more important) the memory usage could not be improved because the final message would always be calculated during linting.

RandomByte:
LGTM and I'm fine with using this for the time being.

I still find the signature somewhat complicated. It would be nice if we could limit it to a few possible parameters, maybe generalizing some of them to fit all possible messages. But I see that this might not be possible given how different some of the messages are.

An alternative to this would be to have dedicated functions per message type, each with statically typed parameters. I don't think this would make it easier for the consuming modules tough.

RandomByte (approved):

PR #224, opened 2024-07-29
test: Add test case for missing detection of enum destructuring

matz3:
Related to https://github.com/SAP/ui5-linter/pull/152/files#r1687576604

PR #261, opened 2024-08-19
fix: Detect deprecated property access in object destructuring

RandomByte:
Follow up of https://github.com/SAP/ui5-linter/pull/224


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -111,6 +123,64 @@ export default class SourceFileLinter {
...
+	}
+
+	analyzeObjectBindingPattern(node: ts.ObjectBindingPattern) {
+		node.elements.forEach((element) => {
+			if (element.kind === ts.SyntaxKind.BindingElement &&
+				element.name.kind === ts.SyntaxKind.Identifier) {
+				this.analyzeIdentifier(element.name);
+			}
+			// TODO: Check whether handling of nested destructuring is required, example:
+			// `const { SomeObject: { SomeOtherObject } } = coreLib;`
```
d3xter666, line 167: This would be quite a complex task and if we need to handle that, maybe a separate backlog should be created.

It's not only a destructuring, but can become quite complex by messing up destructuring, destructuring with renames, destructuring with array spreading and again destructuring...
As far as I remember we covered this in OpenUI5's JSDoc handler, but I don't think there are currently such usages within UI5

d3xter666 (dismissed):

src/linter/ui5Types/SourceFileLinter.ts
RandomByte: Clarified it a bit better in the comment 👍

d3xter666 (approved):

PR #264, opened 2024-08-20
feat: Allow excluding files from UI5 Linter checks

d3xter666:
This change introduces an `--ignore-pattern` CLI option as well as introduces ui5 linter configuration file.

The configuration file is `ui5lint.config.(js|mjs|cjs)` and currently supports only `ignores` property

JIRA: CPOUI5FOUNDATION-782


src/cli/base.ts
```diff
@@ -29,6 +30,11 @@ const lintCommand: FixedCommandModule<object, LinterArg> = {
 	middlewares: [baseMiddleware],
 	builder: function (args: Argv<object>): Argv<LinterArg> {
 		args.usage("Usage: $0 [options]")
+			.option("ignore-pattern", {
+				describe: "Pattern/files that will be ignored for linting. Can be defined also in ui5ling.config.js",
```
RandomByte, line 34:
```suggestion
				describe: "Pattern/files that will be ignored for linting. Can be defined also in ui5linter.config.js",
```

src/linter/linter.ts
```diff
@@ -8,14 +8,43 @@ import posixPath from "node:path/posix";
...
+		callback: (resource) => {
+			if (!miniChecks?.length) {
+				return true;
+			}
+
+			// Minimatch works with FS and relative paths.
+			// So, we need to convert virtual paths to absolute
+			// FS paths and strip the rootDir
+			const resPath = transformVirtualPathToFilePath(
+				resource.getPath(), options.rootDir, "/resources", "test", "/test-resources")
```
RandomByte, line 39: I think the second and the fourth parameter need to be computed based on the root dir like here: https://github.com/SAP/ui5-linter/blob/3dc90f0f6fa8ea9de273e99a0a604c709d9fa065/src/linter/linter.ts#L77-L78

src/linter/linter.ts
```diff
@@ -8,14 +8,43 @@ import posixPath from "node:path/posix";
...
+				.replace(options.rootDir, "");
+
+			return miniChecks.some((check) => !check.match(resPath, true));
```
RandomByte, line 42: Looks like we missed to discuss this as part of CPOUI5FOUNDATION-782: What about un-ignoring files? (compare to https://eslint.org/docs/latest/use/configure/ignore#unignoring-files-and-directories)

RandomByte (changes requested):

src/linter/linter.ts
d3xter666: What about now?
In certain cases, like testing files in tests, the project cannot be determined.

src/linter/linter.ts
d3xter666: Now it can ignore & unignore files.

src/linter/linter.ts
```diff
@@ -1,21 +1,70 @@
...
+	const filteredCollection = createFilterReader({
 		reader: resourceReader,
+		callback: (resource: Resource) => {
+			if (!ignorePattern?.length) {
+				return true;
+			}
+
+			// Minimatch works with FS and relative paths.
+			// So, we need to convert virtual paths to fs
+			const resPath = transformVirtualPathToFilePath(
```
matz3, line 59: this mapping doesn't consider the project namespace. It maps from `/resources/sap/ui/demo/todo/view/App.view.xml` to `webapp/sap/ui/demo/todo/view/App.view.xml`, which is not correct.

d3xter666: Ah, thanks!
Now, it's fixed

src/linter/linter.ts
```diff
@@ -1,21 +1,87 @@
...
+
+	// Patterns must be only relative (to project's root),
+	// otherwise throw an error
+	ignorePattern.forEach((pattern) => {
+		let notNegatedPattern = pattern;
+		if (pattern.startsWith("!")) {
+			notNegatedPattern = pattern.slice(1);
+		}
+
+		if (/\\/g.test(pattern)) {
```
RandomByte, line 60: Not sure whether I should open this can of worms, because I think we have similar code already elsewhere. But it just occurred to me that on UNIX, a file- or directory name can contain backslashes, e.g. `My\ File.js` is a valid file name on macOS.

So I'm not sure whether we should have this check. Maybe it's enough that paths with non-POSIX separators simply won't match anything?

matz3:
Also LGTM, but I think we should add a short "Configuration" section to the README, so that users know what can be configured and how.

src/linter/linter.ts
d3xter666: As we have agreed, removing this validation, until we find a reasonable solution

src/linter/linter.ts
```diff
@@ -1,21 +1,87 @@
...
+			throw Error(`Patterns must be in POSIX format. "${pattern}" defines non-posix path format.`);
+		}
+
+		if (path.isAbsolute(notNegatedPattern)) {
+			throw Error(`Patterns must be relative to project's root folder. "${pattern}" defines an absolute path.`);
```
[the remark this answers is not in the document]
d3xter666: Done

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
...
+Pattern/files that will be ignored for linting. Can be defined also in `ui5lint.config.js`.
+
+**Example:**
+```sh
+ui5lint --ignore-pattern "./test"
+```
+
+## Configuration
+
+UI5 Linter can be easily configured using an external configuration file. This file allows you to customize how the linter behaves, such as ignoring specific files or directories.
```
KlattG, line 109:
```suggestion
The UI5 linter can easily be configured with an external configuration file, allowing you to customize how the linter behaves. For example, you can tell it to ignore specific files or directories. 
```

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
...
+
+## Configuration
+
+UI5 Linter can be easily configured using an external configuration file. This file allows you to customize how the linter behaves, such as ignoring specific files or directories. 
+
+### Configuration File Location
+The configuration file must be placed in the root directory of your project, alongside the `ui5.yaml` and `package.json` files. The linter will automatically detect and load the file when it runs.
+
+### Supported Configuration File Names
+You can name your configuration file in one of the following formats:
```
KlattG, line 115:
```suggestion
You can name your configuration file using one of the following formats:
```

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
...
+  ignores: [
+    "test/**/*", 
+    "!test/sap/m/visual/Wizard.spec.js",
+  ],
+};
+```
+
+### Configuration Options
+
+- **ignores**: This option allows you to define glob patterns to ignore specific files or directories during linting. Patterns are relative to the root of the project. You can also un-ignore specific files by using the `!` prefix. The order of the patterns matters—later patterns can override earlier ones.
```
KlattG, line 146:
```suggestion
- **ignores**: This option allows you to define glob patterns to ignore specific files or directories during linting. Patterns are relative to the root of the project. You can also un-ignore specific files by using the `!` prefix. The order of the patterns matters; later patterns override earlier ones.
```

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
...
+
+Example:
+```js
+ignores: [
+  "test/**/*",              // Ignore all files in the test folder
+  "!test/sap/m/visual/**",  // Un-ignore files in a specific subdirectory
+];
+```
+
+This provides flexible control over which files the UI5 Linter should process and which it should ignore.
```
KlattG, line 156:
```suggestion
In this way, you can control which files the UI5 linter should process and which it should ignore.
```

src/cli/base.ts
```diff
@@ -30,6 +31,11 @@ const lintCommand: FixedCommandModule<object, LinterArg> = {
 	middlewares: [baseMiddleware],
 	builder: function (args: Argv<object>): Argv<LinterArg> {
 		args.usage("Usage: $0 [options]")
+			.option("ignore-pattern", {
+				describe: "Pattern/files that will be ignored for linting. Can be defined also in ui5linter.config.js",
```
KlattG, line 35:
```suggestion
				describe: "Pattern/files that will be ignored during linting. Can also be defined in ui5linter.config.js",
```

KlattG (changes requested):

src/utils/ConfigManager.ts
```diff
@@ -0,0 +1,58 @@
...
+		let config = {} as UI5LintConfigType;
+
+		if (this.#configFile) {
+			// Relative paths are needed to make it work on Windows
+			const configFilePath = this.#resolveModulePaths(this.#configFile);
+			({default: config} = await import(configFilePath) as {default: UI5LintConfigType});
+		} else {
+			// Find configuration file
+			({default: config} = await Promise.any(
+				CONFIG_FILENAMES.map(
```
matz3, line 46: Not sure whether this works, but can't we just import `ui5lint.config` and let Node.ja decide on the extension?

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
 ui5lint --format json
 ```
 
+#### `--ignore-pattern`
+
+Pattern/files that will be ignored for linting. Can be defined also in `ui5lint.config.js`.
+
+**Example:**
+```sh
+ui5lint --ignore-pattern "./test"
```
matz3, line 104: Maybe use a more realistic pattern, such as `webapp/thirdparty/**`?

README.md
```diff
@@ -95,6 +95,66 @@ Choose the output format. Currently, `stylish` (default), `json` and `markdown`
...
+If you need to specify a custom configuration file, you can provide it using the `--config` parameter via the command line.
+
+### Configuration File Format
+
+#### ESM (ECMAScript Modules):
+```js
+export default {
+  ignores: [
+    "test/**/*", 
+    "!test/sap/m/visual/Wizard.spec.js",
```
matz3, line 129: Same here: Rather use an example that is more common for users (which usually have application projects)

README.md
d3xter666: Thanks!

README.md
d3xter666: Now should be better :)

src/utils/ConfigManager.ts
d3xter666: The issue is that if we don't provide file extension, it cannot resolve the configurations with `mjs` and `cjs` extensions.

src/utils/ConfigManager.ts
```diff
@@ -0,0 +1,58 @@
...
+					(filename) => {
+						// Relative paths are needed to make it work on Windows
+						const configFilePath = this.#resolveModulePaths(filename);
+						return import(configFilePath) as Promise<{default: UI5LintConfigType}>;
+					}))
+				// Promise.any would throw if nothing is found i.e. there's no config file
+				.catch(() => ({default: {}})));
```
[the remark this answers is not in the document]
d3xter666: Thanks!
Now, if it's not a module not found exception, it's propagated

src/linter/linter.ts
```diff
@@ -273,3 +280,90 @@ async function fileExists(dirPath: string) {
...
+/**
+ * Function to check if a file path matches the patterns
+*/
+function isFileIncluded(file: string, patterns: string[]) {
+	let include = true;
+
+	for (const pattern of patterns) {
+		if (pattern.startsWith("!")) {
+			// Handle negation: exclude if the file matches the pattern
+			if (minimatch(file, pattern.slice(1))) {
```
RandomByte, line 293: Since the patterns stay the same same for all files, I think it make sense to create one minimatch instance per pattern upfront and have them in an array to go through here

Especially for larger patterns this can save us some time. You can tell negative patterns apart from the others using the [`negate`](https://github.com/isaacs/minimatch?tab=readme-ov-file#properties) property. And use the option [`{nonegate: true}`](https://github.com/isaacs/minimatch?tab=readme-ov-file#nonegate) to still have the behavior where it returns true if the (negated) pattern matches as currently archived using the `slice(1)` call

RandomByte (changes requested):

RandomByte:
Rebased onto main to resolve conflicts

src/linter/linter.ts
d3xter666: Thanks!
Now it's been optimized. Actually, the flipNegate flag did the work

RandomByte (approved):

KlattG (commented):
LGTM

PR #283, opened 2024-09-05
feat: Add custom config CLI option (--config)

maxreichmann:
JIRA: CPOUI5FOUNDATION-887


src/utils/ConfigManager.ts
```diff
@@ -38,7 +38,11 @@ export default class ConfigManager {
 
 		if (this.#configFile) {
 			// Relative paths are needed to make it work on Windows
-			const configFilePath = this.#resolveModulePaths(this.#configFile);
+
+			const configFilePath = path.isAbsolute(this.#configFile) ?
+				this.#configFile : // If it's an absolute path, transform to POSIX format
```
d3xter666, line 43: This one allows inclusion by absolute path that might not be within the project itself

test/lib/utils/configManager.ts
```diff
@@ -10,8 +16,8 @@ test("Check config file", async (t) => {
 
 	t.deepEqual(config, {
 		ignores: [
+			"!test/sap/m/visual/Magician.spec.js",
 			"test/**/*",
```
matz3, line 20: Usually the exclude comes first (`test/**/*`), and after that some files can be re-included (`!test/sap/m/visual/Magician.spec.js`). Note that the negation in this case does not mean "ignore", but "not ignore" (i.e. include).

The re-include in this case should not have any effect as the exclude comes afterwards.

src/utils/ConfigManager.ts
maxreichmann: Not planned

test/lib/utils/configManager.ts
RandomByte: Done

RandomByte (approved):
LGTM

src/utils/ConfigManager.ts
RandomByte: Config files may be located anywhere on the system if the path is provided via CLI. Just like for the ui5.yaml in case of UI5 Tooling

PR #284, opened 2024-09-06
fix: Ensure that results are in deterministic order

matz3:
Previously, only some formatters would sort the results by file path.
This could cause non-deterministic output (e.g. in json format) as some
processing happens asynchronously in parallel.

Sorting of messages within a file is still done by the formatters as
they use different sorting criteria. However, the order of message is
already expected to be deterministic.


test/lib/linter/snapshots/linter.ts.md
```diff
@@ -821,6 +804,23 @@ Generated by [AVA](https://avajs.dev).
         ],
         warningCount: 0,
       },
+      {
+        coverageInfo: [],
+        errorCount: 1,
+        fatalErrorCount: 0,
+        filePath: 'ui5.yaml',
```
matz3, line 811: This one confuses me. I'll have to check what happens here. I would have expected that `ui5.yaml` comes before `webapp/`, as `u` comes before `w`.

matz3: Solved in 9669083

d3xter666 (approved):
LGTM

PR #301, opened 2024-09-13
refactor(messages): Remove '-js' suffix from 'ui5-linter-no-globals-js'

RandomByte:
This rule is used for JS as well as XML files, so the js suffix may be
misleading.


d3xter666 (dismissed):
LGTM

matz3 (approved):

matz3 (approved):

Issue #305, opened 2024-09-15
Unexpected or inconsistent reports of globals


codeworrior:
## Expected Behavior

When a global symbol is used in code, either each usage should be reported, or none. Global APIs that are part of the platform (ES2023, browser) should never be reported. Also, non-UI5 globals, whose usage is legitimated via ESLint config maybe should not be reported (to be discussed).

## Current Behavior

When linting the OpenUI5 sources, platform globals are reported unexpectedly:
* src/sap/ui/core/Component.js
  * [224:18](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/core/Component.js#L224): Access of global variable 'Symbol' (Symbol)
* src/sap/ui/core/date/UI5Date.js
  * [139:20](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/core/date/UI5Date.js#L139): Access of global variable 'Symbol' (Symbol.toStringTag)

In OpenUI5's own test code, `QUnit` is most of the time used as a global (which is accepted in UI5 code). But those usages are sometimes reported (seems, only when `QUnit` is accessed via indexed member access `QUnit[...something]`):
* test/sap/ui/core/qunit/loader/asyncMode.qunit.js
  * [277:3](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/test/sap/ui/core/qunit/loader/asyncMode.qunit.js#L277): Access of global variable 'QUnit' (QUnit)
* test/sap/ui/core/qunit/odata/v4/ODataModel.realOData.qunit.js
  * [1008:3](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/test/sap/ui/core/qunit/odata/v4/ODataModel.realOData.qunit.js#L1008): Access of global variable 'QUnit' (QUnit)
  
All other usages are not reported. Either those few usages should also not be reported (e.g. when an eslint comment `/* global QUnit */` or ESLint config exists) or all usages should be reported consistently (not preferred).

## Steps to Reproduce the Issue

1. Lint `sap.ui.core` from OpenUI5 or, if you have access to internal builds, look at the nightly UI5 2.0 transformation build and check the linting result file `ui5lint-openui5-_openui5_sap.ui.core.json`. 

## Context

- UI5 linter version: **`0.3.4`**
- Node.js Version: **`20.11.1`**
- npm Version: `10.2.4`

## Log Output / Stack Trace

See findings above.

PR #308, opened 2024-09-16
fix(NoGlobals): Fix false-positives for ElementAccessExpressions and some built-in globals

RandomByte:
Resolves https://github.com/SAP/ui5-linter/issues/305


codeworrior (approved):
looks good

PR #312, opened 2024-09-19
feat: Allow usage of patterns in UI5 linter's file-paths config

d3xter666:
JIRA: CPOUI5FOUNDATION-881


src/linter/linter.ts
```diff
@@ -11,8 +11,47 @@ import type {AbstractReader, Resource} from "@ui5/fs";
...
+	resourceReader: AbstractReader, options: LinterOptions, config: UI5LintConfigType
+): Promise<LintResult[]> {
+	const lintEnd = taskStart("Linting");
+	let {ignorePattern, filePatterns} = options;
+	const {rootDir} = options;
+
+	// Resolve files to include
+	filePatterns = [
+		...(config.files ?? []),
+		...(filePatterns ?? []), // CLI patterns go after config patterns
```
RandomByte, line 24: I'm not sure whether this is the behavior a user would expect. If the config file specifies `webapp/**/*`, and the user provides `webapp/controller/BaseController.js` via the CLI, the current implementation will still lint everything in `webapp`. I think the user would rather expect to only lint `webapp/controller/BaseController.js`.

What do you think? Maybe we should ignore the config file if filePatterns are provided?

d3xter666: Yep, makes totally sense.
Maybe, I have been confused by the ignores property where we merge them, but the CLI has precedence.
I'll refactor it!

Thanks!

src/cli/base.ts
```diff
@@ -21,18 +22,22 @@ export interface LinterArg {
...
+	command: "$0 [files...]",
 	describe: "Runs linter",
 	handler: handleLint,
 	middlewares: [baseMiddleware],
 	builder: function (args: Argv<object>): Argv<LinterArg> {
-		args.usage("Usage: $0 [options]")
+		args.usage("Usage: $0 [files...] [options]")
+			.positional("files", {
+				describe: "List of patterns to lint",
+				type: "string",
```
RandomByte, line 39: I think this is missing `array: true` (see https://github.com/SAP/ui5-cli/blob/d33c3ab8a63345e199d731a0895ec8db9a32abde/lib/cli/commands/build.js#L42)

However, even then it doesn't seem to transform single positionals into an array. 

Executing `ui5lint --files webapp/controller/App.controller.js` in the troublesome app yields the following error:
```
Error Message:
patterns.map is not a function

Stack Trace:
TypeError: patterns.map is not a function
    at buildPatterns (/Users/d054485/SAPDevelop/dev/fnd/ui5-linter/src/linter/linter.ts:278:18)
    at resolveReader (/Users/d054485/SAPDevelop/dev/fnd/ui5-linter/src/linter/linter.ts:337:28)
```

RandomByte (changes requested):

src/cli/base.ts
d3xter666: It should be `ui5lint webapp/controller/App.controller.js`

The `--files` option I suppose, comes from `yargs`. I have updated the code to enforce array for that.

RandomByte (approved):

PR #316, opened 2024-09-23
feat(html): Detect deprecated bootstrap parameters

RandomByte:
JIRA: CPOUI5FOUNDATION-906

PR #320, opened 2024-09-24
feat: Detect deprecated View file types

maxreichmann:
JIRA: CPOUI5FOUNDATION-883


src/linter/messages.ts
```diff
@@ -175,6 +178,25 @@ export const MESSAGE_INFO = {
 		details: ({details}: {details: string}) => details,
 	},
 
+	[MESSAGE.DEPRECATED_VIEW_CONFIG]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-deprecated-api"],
```
d3xter666, line 183: I know we're trying to consolidate those, but I'm still wondering whether we need to distinguish between API deprecation, partial deprecation and whole files/patterns deprecation...

Just food for thought and something we might want to discuss.
My point is that in the future we might want to do some filtering based on ruleId and `no-deprecated-api` would have a significant percentage of the findings

src/linter/fileTypes/linter.ts
```diff
@@ -0,0 +1,39 @@
...
+	".view.js",
+	".view.tmpl",
+	".view.ts",
+	".fragment.html",
+	".fragment.js",
+	".fragment.ts",
+];
+
+export default async function lintFileTypes({workspace, context}: LinterParameters) {
+	let xmlResources: Resource[];
```
d3xter666, line 17: To me this var name seems a bit ambiguous as it refers to XMLResources, but what we're looking for is anything, but XML :D 

Maybe, `potentialDeprecatedResources` or something in that direction

d3xter666, line 17: To me this var name seems a bit ambiguous as it refers to XMLResources, but what we're looking for is anything, but XML :D 

Maybe, `potentialDeprecatedResources` or something in that direction

src/linter/messages.ts
RandomByte: Yes, absolutely. Similarly in https://github.com/SAP/ui5-linter/pull/316

I find it hard to come up with rule names that generalize some of our messages without getting too specific again. Ultimately most of these findings are reported because something got "deprecated", but it's not really an API. So "no-deprecated-view-type"? That's already almost too specific and can only cover few of the messages.

We should look at the whole list we have now and re-group things ideally before the 1.0 release.

d3xter666 (approved):
LGTM with some small remarks

src/linter/fileTypes/linter.ts
maxreichmann: Duplicate comment.

maxreichmann: Done

src/linter/messages.ts
RandomByte: FW, FV and me discussed this briefly. Looking at it from an app developer perspective it seems reasonable to collect all deprecations under a single (or very few) rules. We couldn't identify any benefit from differentiating between "deprecated views", "deprecated properties", "deprecated parameters" etc.

Ultimately a developer will have to deal with all of these cases very similarly.

d3xter666 (approved):
LGTM

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -119,4 +165,15 @@ export default class ManifestLinter {
 			}
 		});
 	}
+
+	detectDeprecatedPropertyNames(viewObject: Record<string, unknown>, pathToViewObject: string) {
+		// Detect every property name starting with "view" except "viewType":
+		for (const key of Object.keys(viewObject)) {
+			if (key.startsWith("view") && key !== "viewType") {
+				this.#reporter?.addMessage(MESSAGE.DEPRECATED_VIEW_CONFIG, {
```
matz3, line 173: This does seem to produce some false-positive findings that cause issues when changing "viewName" to "name" in routing "targets". The application won't come up at all without an error unless sap.ui5/routing/config does specify `"type": "View"`.

We should improve our linting check and ideally the routing should log an error to explain why nothing is rendered.

cc @stopcoder

stopcoder: The `type` option is mandatory when defining `Targets` in the routing config. This option is introduced with the new option set in routing. I think we need to enhance the hint by adding the information when options with prefix "view" are found saying that the "type": "View" needs to be added as well.

matz3: Thanks for the insights. But what about the case when options are not prefixed? Isn't the "type" then also mandatory? With the view prefix it works without the "type".

stopcoder: No. `type` is only needed for the new option set (the none-prefixed ones). `type` can also be provided in parallel to the prefixed options but it doesn't have any effect and routing doesn't complain about it.

matz3: Yes, that's what I was saying. If the new option set is used without `type`, then nothing happens and the app does not render.

matz3: See https://github.com/SAP/ui5-linter/pull/356

Issue #332, opened 2024-10-01
README outdated after v1.0.0


xLexip:
Hey there,

the `--file-paths `option is still in the README, even though it was removed with **v1.0.0**. 
https://github.com/SAP/ui5-linter?tab=readme-ov-file#--file-paths

Additional, it would be nice to explicitly list the change from `--file-paths` to the _files_ positional as a breaking change in the release notes. Our pipelines ran with Anlauf into error. :)

Best regards,
Lexip

RandomByte:
Many thanks for letting us know! Indeed we missed that update. Will be fixed soon.

PR #333, opened 2024-10-01
docs: Update documentation

d3xter666:
fixes: https://github.com/SAP/ui5-linter/issues/332

JIRA: CPOUI5FOUNDATION-881

Update documentation after the implementation of https://github.com/SAP/ui5-linter/pull/312

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ UI5 linter report:
 Note: Use "ui5lint --details" to show more information about the findings
 ```
 
-You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results.
+You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results. Note: this can't be used to include files that normally wouldn't be checked (e.g. files outside of `webapp` in application projects)
 
 ```sh
 > ui5lint webapp/**/*
--- a/README.md
+++ b/README.md
@@ -80,14 +80,15 @@ You can provide multiple glob patterns as arguments after the `ui5lint` command 
 **Note**: Only POSIX separators are allowed, regardless of the target platform.
 
 ```sh
-> ui5lint "webapp/**/*.xml"
+> ui5lint "application/webapp/**/*.xml"
 
 UI5 linter report:
 
-webapp/view/App.view.xml
-  7:8 error Import of deprecated module 'sap/f/Avatar'
+/application/webapp/view/Main.view.xml
+  16:39 error Import of deprecated module 'sap/m/MessagePage'
+  22:5  error Use of deprecated property 'blocked' of class 'Button'
 
-1 problems (1 errors, 0 warnings)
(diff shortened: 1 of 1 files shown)


KlattG (approved):
LGTM

README.md
```diff
@@ -73,17 +73,45 @@ UI5 linter report:
 Note: Use "ui5lint --details" to show more information about the findings
 ```
 
-### Options
+You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results.
 
-#### `--file-paths`
+```sh
+> ui5lint webapp/**/*
```
RandomByte, line 79:
```suggestion
> ui5lint 'webapp/**/*'
```

Using `webapp/**/*` will expand the glob in the shell (before ui5lint is executed). The linter will receive a list of all files matching the pattern as an argument

Using `'webapp/**/*` will pass the pattern as an argument to ui5lint, which then passes it to _our_ glob function.

I'm not entirely sure which way is better, but I think the latter is more likely to behave the same across platforms and shells, so maybe we should document that. @codeworrior do you have an opinion on this?

RandomByte: Maybe it makes sense to show an example where less files than usual are linted? Something like `webapp/**/*.xml` to only check XMLs maybe? Then report example below needs to be updated as well in that case.

README.md
```diff
@@ -73,17 +73,45 @@ UI5 linter report:
 Note: Use "ui5lint --details" to show more information about the findings
 ```
 
-### Options
+You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results.
```
RandomByte, line 76:
```suggestion
You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results. Note: this can't be used to include files that normally wouldn't be checked (e.g. files outside of `webapp` in application projects)
```

RandomByte (changes requested):

README.md
d3xter666: Docs are updated :)

RandomByte (dismissed):
LGTM. @KlattG would you mind taking another look? We changed some of the texts.

README.md
```diff
@@ -73,17 +73,30 @@ UI5 linter report:
 Note: Use "ui5lint --details" to show more information about the findings
 ```
 
-### Options
+You can provide multiple glob patterns as arguments after the `ui5lint` command to filter and narrow down the linting results. 
 
-#### `--file-paths`
+**Note**: this can't be used to include files that normally wouldn't be checked (e.g. files outside of `webapp` in application projects)
```
KlattG, line 78:
```suggestion
**Note**: This option does not permit you to include files that normally wouldn't be checked (e.g. files outside of the `webapp` folder in application projects).
```

KlattG (changes requested):
Just a small suggestion

flovogt (approved):

PR #334, opened 2024-10-01
fix: Check unmatched patterns

d3xter666:
JIRA: CPOUI5FOUNDATION-881


src/linter/linter.ts
```diff
@@ -351,3 +360,33 @@ export async function resolveReader({
...
 	});
 }
+
+/**
+ * Stash patterns
+ *
+ * @param pattern Pattern to stash as matched
+ */
+function patternMatched(pattern: FilePattern) {
+	patternsMatch[pattern] = patternsMatch[pattern] ? patternsMatch[pattern] + 1 : 1;
```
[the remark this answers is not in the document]
d3xter666: Currently, not, but I was wondering whether this could serve us somehow...
I'll just use the Set. Thanks! It'd be quicker and simpler :)

src/linter/linter.ts
```diff
@@ -346,8 +360,29 @@ export async function resolveReader({
...
+		if (!patternsMatch.has(pattern)) {
+			acc.push(pattern);
+		}
+
+		return acc;
+	}, [] as FilePattern[]);
+
+	if (unmatchedPatterns.length) {
+		throw new Error(`${unmatchedPatterns.length === 1 ? "Pattern" : "Patterns"}` +
+			` '${unmatchedPatterns.join("', '")}' did not match any resource!`);
```
RandomByte, line 386:
```suggestion
		throw new Error(`Specified file ${unmatchedPatterns.length === 1 ? "pattern" : "patterns"}` +
			` '${unmatchedPatterns.join("', '")}' did not match any resource`);
```

RandomByte (changes requested):

RandomByte (approved):

src/linter/linter.ts
```diff
@@ -11,6 +11,11 @@ import type {AbstractReader, Resource} from "@ui5/fs";
 import ConfigManager, {UI5LintConfigType} from "../utils/ConfigManager.js";
 import {Minimatch} from "minimatch";
 
+// Internal analytical variable that will store matchers' count.
+// We cannot predict the outcome of the matchers, so stash usage statistics
+// and later analyze the results from it.
+const matchedPatterns = new Set<string>();
```
matz3, line 17: Having this here as a singleton set will most probably cause issues when running multiple linting activities in the same process in parallel. We should think about moving it to an instance of the current linting activity, which might require some refactoring (e.g. turning this module into a class might help).

PR #349, opened 2024-10-07
feat: Detect deprecations in ManagedObject metadata

d3xter666:
JIRA: CPOUI5FOUNDATION-858

Depends on: https://github.com/SAP/ui5-linter/pull/358


scripts/metadataProvider/createDeprecationsInfo.ts
```diff
@@ -0,0 +1,59 @@
...
+	const apiExtract = {
+		framework: {
+			name: "SAPUI5",
+			version: sapui5Version,
+		},
+		deprecations,
+	};
+
+	await writeFile(
+		new URL("../../resources/api-deprecations.json", import.meta.url),
```
matz3, line 56: what about adding this information to the existing `api-extract.json` instead of creating another file?
The idea of the `api-extract.json` was to extract all necessary information for linting from the `api.json` files, which explains its name.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -120,12 +124,154 @@ export default class SourceFileLinter {
...
+			metadataFound();
+			return false;
+		}
+
+		while (node && node.kind !== ts.SyntaxKind.ClassDeclaration) {
+			node = node.parent;
+		}
+
+		if (!node) {
+			return false;
```
matz3, line 184: missing call to `metadataFound();` ?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -120,12 +124,154 @@ export default class SourceFileLinter {
 				context: this.#context,
 				checker: this.#checker,
 			});
+		} else if (ts.isPropertyAssignment(node) && this.isPropertyInMetadata(node)) {
+			this.analyzeMetadataProperty(node.name.getText(), node);
```
matz3, line 128: Do I understand this right that we run the `this.isPropertyInMetadata(node)` check for every single property within the MO metadata (i.e. multiple times for the same class)?

I would hope that this can be solved differently, so that we only need to check for this once. Maybe the entry for the check can be done at the ClassDeclaration and then we know that all properties below that node are part of the class. As we plan other checks as well like checking for renderer of controls (which are also sub-classes of ManagedObject), we should try to only check for the inheritance chain at a single place and then run checks based on the class type.

test/fixtures/linter/rules/NoDeprecatedApi/ManagedObject.js
```diff
@@ -0,0 +1,78 @@
+sap.ui.define(["sap/ui/core/Control", "sap/m/library", "mycustom/lib/FancyMultiPage"],
```
matz3, line 1: The file is called ManagedObject, but the class inherits from Control. I think it makes sense to test both: A sub-class of ManagedObject and a sub-class of Control.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Indeed, I was wondering what is the best approach to do so.
That's why I put those pef measurements. Basically, for every prop, we check whether it's a `metadata` child and if the control is child of ManagedObject.

The tricky part here is that nodes are visited recursivelly in depth-first approach. And I'm not sure how to handle the check that the current node is actually inside the `metadata` property.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks!

scripts/metadataProvider/createDeprecationsInfo.ts
d3xter666: Done

test/fixtures/linter/rules/NoDeprecatedApi/ManagedObject.js
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: These are the performance results for the following check:
```js
if (ts.isPropertyAssignment(node) && this.isPropertyInMetadata(node)) {
```

### Base- no check at all
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `node /Users/i326076/SAPDevelop/ui5-linter/bin/ui5lint.js` | 39.019 ± 0.422 | 38.311 | 39.921 | 1.00 |


### Check- just the check, `this.analyzeMetadataProperty(node.name.getText(), node);` is not executed.
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `node /Users/i326076/SAPDevelop/ui5-linter/bin/ui5lint.js` | 39.006 ± 0.229 | 38.657 | 39.338 | 1.00 |

d3xter666: This is the alternative implementation. Still polishing it, but I have doubts it will be quicker as it will be hard to optimise the traversal, so metdata nodes to be visited only once: https://github.com/SAP/ui5-linter/pull/360

d3xter666: WIth https://github.com/SAP/ui5-linter/pull/360 merged in this branch the measurements are alike the main branch

matz3 (dismissed):

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -120,12 +124,156 @@ export default class SourceFileLinter {
 				context: this.#context,
 				checker: this.#checker,
 			});
+		} else if ((ts.isPropertyAssignment(node) || // Control.extend("sap.fancy.Text", {metadata: {....}})
```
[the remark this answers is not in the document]
d3xter666: Thanks!

test/fixtures/linter/rules/NoDeprecatedApi/Control_ManagedObject.js
```diff
@@ -0,0 +1,78 @@
...
+				eventC: {
+					parameters: {
+						layout: "Promise<sap.f.AvatarShape>", // deprecated
+					},
+				},
+			},
+		},
+	});
+
+	// FancyText.prototype._getBeginColumn = function () {
```
matz3, line 68: Can be removed?

test/fixtures/linter/rules/NoDeprecatedApi/ManagedObject.js
```diff
@@ -0,0 +1,78 @@
...
+				eventC: {
+					parameters: {
+						layout: "Promise<sap.f.AvatarShape>", // deprecated
+					},
+				},
+			},
+		},
+	});
+
+	// FancyText.prototype._getBeginColumn = function () {
```
matz3, line 68: Can be removed?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -120,12 +124,156 @@ export default class SourceFileLinter {
...
+		// This one is too generic and should always be at the last place
+		// It's for "types" and event arguments' types
+		} else if (ts.isStringLiteral(node.initializer)) {
+			const deprecatedTypes = {
+				...this.#apiExtract.deprecations.enums,
+				...this.#apiExtract.deprecations.typedefs,
+			} as Record<string, string>;
+
+			let nodeType = node.initializer.text;
+			nodeType = nodeType.replace("Promise<", "").replace(">", ""); // Cleanup event types
```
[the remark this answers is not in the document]
matz3: Looks good. A test case e.g. for `sap.m.P13nColumnsItem[]` would be good to make sure this is covered.

test/fixtures/linter/rules/NoDeprecatedApi/Control_ManagedObject.js
```diff
@@ -0,0 +1,67 @@
...
+				// sap.m.DateTimeInputType: deprecated, "Date": deprecated
+				textDirectionB: {type: "sap.m.DateTimeInputType", group: "Appearance", defaultValue: "Date"},
+				// sap.m.FrameType: NOT deprecated, sap.m.FrameType.TwoThirds: deprecated
+				textAlign: {type: "sap.m.FrameType", group: "Appearance", defaultValue: FrameType.TwoThirds},
+				// sap.m.FrameType: NOT deprecated, "TwoThirds": deprecated
+				textAlignB: {type: "sap.m.FrameType", group: "Appearance", defaultValue: "TwoThirds"},
+			},
+			aggregations: {
+				// ap.m.P13nColumnsItem: deprecated
+				myagg: {type: " sap.m.P13nColumnsItem[]", multiple: false, visibility: "hiddenDeprecated"},
```
matz3, line 24: I don't think that this definition is correct. The aggregation is not even typed as "multiple".
The cases I could find were mainly using the array notation within event parameters, see https://github.com/SAP/openui5/blob/82b2165a325f8ff2f5388195e176929f39ded271/src/sap.m/src/sap/m/P13nColumnsPanel.js#L151

d3xter666: Thanks!

matz3 (commented):
LGTM

matz3 (approved):

PR #365, opened 2024-10-17
fix: False positive for deprecated JS view/fragment type

matz3:
The *.view.js / *.fragment.js file naming is just a convention but does
not imply the usage of deprecated view/fragment API.
As JS/TS files are checked already and usages of sap.ui.jsview /
sap.ui.jsfragment are reported, there is no need to implement a separate
check.


codeworrior (approved):
LGTM, thanks!

PR #366, opened 2024-10-17
feat: Detect render deprecations embed in Control

d3xter666:
JIRA: CPOUI5FOUNDATION-878

### DIFF
--- a/src/linter/messages.ts
+++ b/src/linter/messages.ts
@@ -45,6 +45,7 @@ export enum MESSAGE {
 	HTML_IN_XML,
 	LIB_INIT_API_VERSION,
 	MISSING_BOOTSTRAP_PARAM,
+	NO_DEPRECATED_RENDERER,
 	NO_DIRECT_DATATYPE_ACCESS,
 	NO_DIRECT_ENUM_ACCESS,
 	NO_GLOBALS,
@@ -307,6 +308,15 @@ export const MESSAGE_INFO = {
 		message: ({libInitFunction}: {libInitFunction: string}) =>
 			`Deprecated call to ${libInitFunction}(). Use the {apiVersion: 2} parameter instead`,
 		details: () => `{@link sap.ui.core.Lib.init Lib.init}`,
+	},
+
+	[MESSAGE.NO_DEPRECATED_RENDERER]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-deprecated-api"],
+
+		message: () =>
+			`Use of deprecated renderer. Define explicitly {apiVersion: 2} parameter in renderer object`,
+		details: () => `"{@link topic:c9ab34570cc14ea5ab72a6d1a4a03e3f Renderer Object}",`,
 	},
 
 	[MESSAGE.NO_DIRECT_DATATYPE_ACCESS]: {
--- a/src/linter/ui5Types/SourceFileLinter.ts
+++ b/src/linter/ui5Types/SourceFileLinter.ts
@@ -136,10 +136,45 @@ export default class SourceFileLinter {
 			if (this.isUi5controlMetadataNode(node)) {
 				ts.forEachChild(node, visitMetadataNodes);
 			}
+		} else if ((ts.isPropertyAssignment(node) || ts.isPropertyDeclaration(node) ||
+			ts.isMethodDeclaration(node) || ts.isArrowFunction(node)) &&
+			["renderer", "render"].includes(node.name?.getText())) {
+			this.analyzeControlRenderer(node);
 		}
 
 		// Traverse the whole AST from top to bottom
 		ts.forEachChild(node, this.#boundVisitNode);
+	}
+
+	analyzeControlRenderer(node: ts.PropertyAssignment |
+		ts.PropertyDeclaration | ts.MethodDeclaration | ts.ArrowFunction) {
+		if ((ts.isPropertyAssignment(node) || ts.isPropertyDeclaration(node)) &&
+			node.initializer && ts.isObjectLiteralExpression(node.initializer)) {
+			const apiVersionNode = node.initializer?.properties.find((prop) => {
+				return ts.isPropertyAssignment(prop) &&
+					ts.isIdentifier(prop.name) &&
+					prop.name.text === "apiVersion";
+			});
+
+			let nodeToHighlight: ts.PropertyAssignment | ts.PropertyDeclaration | undefined = undefined;
+			if (!apiVersionNode) { // No 'apiVersion' property
+				nodeToHighlight = node;
+			} else if (ts.isPropertyAssignment(apiVersionNode) &&
+				apiVersionNode.initializer.getText() !== "2") { // String value would be "\"2\""
+				nodeToHighlight = apiVersionNode;
+			}
+
+			if (nodeToHighlight) {
+				this.#reporter.addMessage(MESSAGE.NO_DEPRECATED_RENDERER, nodeToHighlight);
+			}
+		} else if ((ts.isPropertyAssignment(node) || ts.isPropertyDeclaration(node)) && node.initializer &&
+			(ts.isMethodDeclaration(node.initializer) || ts.isArrowFunction(node.initializer)) &&
+			node.name?.getText() === "renderer") {
+			this.#reporter.addMessage(MESSAGE.NO_DEPRECATED_RENDERER, node);
+		} else if ((ts.isPropertyAssignment(node) || ts.isPropertyDeclaration(node)) && node.initializer &&
+			ts.isIdentifier(node.initializer)) {
+			// TODO: detect renderer variable that's been assigned a function
+		}
 	}
 
 	isUi5controlMetadataNode(node: ts.PropertyAssignment | ts.PropertyDeclaration): boolean {


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -136,12 +136,91 @@ export default class SourceFileLinter {
...
+				node.initializer && ts.isIdentifier(node.initializer)) ||
+				ts.isShorthandPropertyAssignment(node)) {
+			let renderInitializer = ts.isShorthandPropertyAssignment(node) ?
+				node.name.getText() :
+				node.initializer!.getText();
+
+			let useSource = false;
+			const sourceFile = node.getSourceFile();
+
+			const findRenderDeclaration = (potentialDeclarations: ts.Node) => {
```
matz3, line 194: What about using the typeChecker (`this.#checker.getTypeAtLocation`) instead to resolve the type?

matz3: Also see: https://github.com/SAP/ui5-linter/pull/374/files

RandomByte:
LGTM

test/fixtures/linter/rules/Render/2ControlRenderer.js
```diff
@@ -0,0 +1,11 @@
+// This file would be skipped from direct testing.
```
RandomByte, line 1: Could we add a test where only the renderer is linted? I guess there wouldn't be any findings, meaning that one has to always lint the Control together with the renderer to see all findings in the renderer?

matz3: Another scenario would be where the renderer is ignored. I wonder whether there would still be a message, as it is created through linting of a not ignored file (control).

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,203 @@ export default class SourceFileLinter {
...
+				declarations?.forEach((declaration) => this.analyzeControlRendererInternals(declaration));
+			} else {
+				// Analyze renderer property when it's directly embedded in the renderer object
+				// i.e. { renderer: {apiVersion: "2", render: () => {}} }
+				this.analyzeControlRendererInternals(rendererMember.initializer);
+			}
+		} else if (ts.isShorthandPropertyAssignment(rendererMember)) {
+			// Special case for shorthand property assignment. Basically the same as the Identifier case above
+			const {symbol: {declarations}} = this.#checker.getTypeAtLocation(rendererMember);
+			declarations?.forEach((declaration) => this.analyzeControlRendererInternals(declaration));
```
matz3, line 267: This case is currently not covered by the tests

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,203 @@ export default class SourceFileLinter {
...
+			const apiVersionNode = node.properties.find((prop) => {
+				return ts.isPropertyAssignment(prop) &&
+					ts.isIdentifier(prop.name) &&
+					prop.name.text === "apiVersion";
+			});
+
+			let nodeToHighlight: ts.PropertyAssignment | ts.PropertyDeclaration |
+				ts.VariableDeclaration | ts.ObjectLiteralExpression | undefined = undefined;
+			if (!apiVersionNode) { // No 'apiVersion' property
+				nodeToHighlight = node;
```
matz3, line 284: Line not covered

test/fixtures/linter/rules/Render/2ControlRenderer.js
d3xter666: Thanks! I would need to update the message. My initial intention was to include only the Controls, and skip the Renders being checked in tests, but this seemed to be a wrong concept.

Now, all the files are checked. The special case is that Renders are checked twice- once "normally" as we do with the rest of the file and ones through the Controls' renderer context.

Maybe, I need to add some general deprecation to prove this, like a global variable reference

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: This one is a bit hard to test now with the improved extend call transpilation, because of that line here:
https://github.com/SAP/ui5-linter/blob/31432b050a5b4a1dc446923331efa11b97652ef2/src/linter/ui5Types/amdTranspiler/rewriteExtendCall.ts#L67

I'll try to create a PR with a better handling of `shorthandPropertyAssignment`s

d3xter666: https://github.com/SAP/ui5-linter/pull/385

d3xter666:
Currently, the tests would fail until https://github.com/SAP/ui5-linter/pull/385 is merged or we decide to postopne shorthand assignment handling for now

test/fixtures/linter/rules/Render/2ControlRenderer.js
RandomByte: I guess this still leaves us with the problem that you get different results for "MyControlRenderer.js" when including or excluding the associated control.

e.g. `ui5lint webapp/controls/MyControl.js webapp/controls/MyControlRenderer.js` vs. `ui5lint webapp/controls/MyControlRenderer.js`. The first command might reveal more issues with the renderer.

But maybe that's good enough for now? CC: @flovogt

d3xter666: Yep, that's true!
And the main reason behind this is that without the binding between the Control and the Renderer, which is defined in the Control, we cannot guarantee that this file (`webapp/controls/MyControlRenderer.js`) is indeed a renderer.
I thinks we already discussed that we are not able to handle this scenario or if we can, it might be too expensive as we need to find the control that this renderer belongs to.

Of course, if there are some general issues in the code itself, like other deprecations, for example, they will be detected.

flovogt: LGTM. If you restrict the scan to a set of specific files you have to know this

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,199 @@ export default class SourceFileLinter {
...
+		// When the render is a plain function
+		if (ts.isMethodDeclaration(node) || ts.isArrowFunction(node) ||
+			ts.isFunctionExpression(node) || ts.isFunctionDeclaration(node)) {
+			renderMethodNode = node;
+		} else if (ts.isObjectLiteralExpression(node)) {
+			// When the render is embed into the renderer object
+			const renderProperty = node.properties.find((prop) => {
+				return ts.isPropertyAssignment(prop) &&
+					ts.isIdentifier(prop.name) &&
+					prop.name.text === "render";
```
codeworrior, line 337: Side note: In SAPUI5, only 13 of 47 `RenderManager#icon` calls are in the `render` method. The others are in helper methods in the renderer.

d3xter666: But is there any way to identify that `oRm.icon(...)` is actually a `RenderManager#icon` call?

codeworrior: 100% safe? No, not without type info. But I guess it would be quite safe to assume a RenderManager#icon call in the context of a Renderer module (if it has been recognised as such) or within the renderer object in a control (embedded renderer scenario).

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,199 @@ export default class SourceFileLinter {
...
 			}
 		}
+
+		if (!renderMethodNode) {
+			return;
+		}
+
+		// If there's a dependency to IconPool from the Renderer,
+		// it's fine and we can skip the rest of the checks
+		const renderManagerSource = renderMethodNode.getSourceFile();
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,199 @@ export default class SourceFileLinter {
...
+		}
+	}
+
+	analyzeControlRendererInternals(node: ts.Node) {
+		// Analyze renderer property when it's an ObjectLiterExpression
+		// i.e. { renderer: {apiVersion: "2", render: () => {}} }
+		if (node && (ts.isObjectLiteralExpression(node) || ts.isVariableDeclaration(node))) {
+			const apiVersionNode = ts.isObjectLiteralExpression(node) && node.properties.find((prop) => {
+				return ts.isPropertyAssignment(prop) &&
+					ts.isIdentifier(prop.name) &&
```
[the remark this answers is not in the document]

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,199 @@ export default class SourceFileLinter {
...
+					prop.name.text === "apiVersion";
+			});
+
+			let nodeToHighlight: ts.PropertyAssignment | ts.PropertyDeclaration |
+				ts.VariableDeclaration | ts.ObjectLiteralExpression | undefined = undefined;
+			if (!apiVersionNode) { // No 'apiVersion' property
+				nodeToHighlight = node;
+			} else if (ts.isPropertyAssignment(apiVersionNode) &&
```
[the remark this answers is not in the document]
d3xter666: Yep, TS detected this as `ts.ObjectLiteralElementLike`.
I have now refactored it with type casting on the assignment and this indeed became redundant

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Now, the detection checks whether `.icon()` call has been made within the `render()` function or within the entire control renderer (if such)

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -219,29 +224,199 @@ export default class SourceFileLinter {
 			return;
 		}
 
-		if (ts.isPropertyDeclaration(rendererMember) && rendererMember.initializer) {
-			const initializerType = this.#checker.getTypeAtLocation(rendererMember.initializer);
+		if ((ts.isPropertyAssignment(rendererMember) || ts.isPropertyDeclaration(rendererMember) ||
+			ts.isVariableDeclaration(rendererMember)) && rendererMember.initializer) {
```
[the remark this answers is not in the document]

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -415,7 +407,18 @@ export default class SourceFileLinter {
...
+		if (rendererSource.fileName !== this.#sourceFile.fileName) {
+			ts.forEachChild(rendererSource, findIconCallExpression);
+		} else {
+			// When the renderer is embedded in the control file, then we can analyze only the icon call
+			// within the render method.
+			ts.forEachChild(renderMethodNode, findIconCallExpression);
+		}
+
+		// Remove duplicate findings
+		// this.#reporter.deduplicateMessages();
```
RandomByte, line 421: Not needed?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -241,9 +246,181 @@ export default class SourceFileLinter {
...
+					}).start);
+			}
+
+			this.analyzeIconCallInRenderMethod(node);
+		// Analyze renderer property when it's a function i.e. { renderer: () => {} }
+		} else if (ts.isMethodDeclaration(node) || ts.isArrowFunction(node) ||
+			ts.isFunctionExpression(node) || ts.isFunctionDeclaration(node)) {
+			// reporter.addMessage() won't work in this case as it's bound to the current analyzed file.
+			// The findings can be in different file i.e. Control being analyzed,
+			// but reporting might be in ControlRenderer
```
matz3, line 304: We should find a better solution to this, especially when there are new checks with the same requirement. For now, this is OK for me and we can tackle it as a follow-up.

test/fixtures/linter/rules/renderer/8ControlRenderer.js
```diff
@@ -0,0 +1,20 @@
+sap.ui.define([], function () {
+	var myControlRenderer = {
+		// apiVersion: 1, // Must be present
```
matz3, line 3: What does `Must be present` mean here? If it means `apiVersion: 1` must be present, I think that wouldn't be correct, as 2 must be present, right?

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Yep 😄 
Thanks!

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: It'd require a greater refactorring as the 
SourceFileReporter is bound to a specific flie as well as the SourceFileLinter itself.

d3xter666:
Hi @KlattG ,

Would you please take a look at the texts in this chenge?
Especially the ones in [src/linter/messages.ts](https://github.com/SAP/ui5-linter/pull/366/files#diff-09701c0232493655601b38f00fc2ebafebdf513a2aaf652ef5d440d1ea3ee04d)

RandomByte (dismissed):

src/linter/messages.ts
```diff
@@ -312,6 +314,24 @@ export const MESSAGE_INFO = {
 		details: () => `{@link sap.ui.core.Lib.init Lib.init}`,
 	},
 
+	[MESSAGE.NO_DEPRECATED_RENDERER]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-deprecated-api"],
+
+		message: () =>
+			`Use of deprecated renderer. Define explicitly {apiVersion: 2} parameter in renderer object`,
```
KlattG, line 322:
```suggestion
			`Use of deprecated renderer detected. Define explicitly the {apiVersion: 2} parameter in the renderer object`,
```

KlattG (changes requested):
Just one suggestion, otherwise fine.

RandomByte (approved):

Issue #355, opened 2024-10-09
UI5 linter does not report findings with given file path pattern


flovogt:
## Expected Behavior

When executing `ui5lint` with a specified file path pattern, it should report findings the same findings as the linter would detect when executed in the specified directory.

`ui5lint my/path/*` and `(cd my/path/ && ui5lint)` should report the same results.

## Current Behavior

`ui5lint test/sap/ui/core/demokit/tutorial/mockserver/01/*` shows an error

````bash
Error Message:
Specified file patterns 'test/sap/ui/core/demokit/tutorial/mockserver/01/Component.js', 'test/sap/ui/core/demokit/tutorial/mockserver/01/manifest.json', 'test/sap/ui/core/demokit/tutorial/mockserver/01/package.json', 'test/sap/ui/core/demokit/tutorial/mockserver/01/ui5.yaml', 'test/sap/ui/core/demokit/tutorial/mockserver/01/webapp' did not match any resource
````

## Steps to Reproduce the Issue

1. Checkout https://github.com/SAP/openui5 and navigate to folder `src/sap.ui.core/`
2. Execute `ui5lint test/sap/ui/core/demokit/tutorial/mockserver/01/*`
3. The reported error is shown
4. When navigating to the folder and execute `(cd test/sap/ui/core/demokit/tutorial/mockserver/01/* ui5lint)` the findings are correctly reported

## Context

- UI5 linter version: **`v1.0.1`**
- Node.js Version: **`v22.9.0`**
- npm Version: `v10.8.9`
- OS/Platform: `macos15`

matz3:
Short update on this one. It indeed seems that there are several unexpected limitations on how files can be specified.
- Specifying a folder without a glob pattern does not lint the contents of that folder
- Specifying any file/pattern in library projects does not seem to work unless the folder namespace is omitted (i.e. `src/library.js`)
- Invoking the CLI in a sub-folder of a project is currently not supported at all

d3xter666:
**Just a note:** The case described in the issue is rather special and we need to make it clear that there will be certain limitations and exceptions. The `test/sap/ui/core/demokit/tutorial/mockserver/01/` path in `sap.ui.core` is a project itself within the `sap.ui.core` library (main project). 
[...]

PR #367, opened 2024-10-17
fix: Correctly handle namespace resolution in linting

d3xter666:
fixes: https://github.com/SAP/ui5-linter/issues/355


src/linter/linter.ts
```diff
@@ -324,17 +328,18 @@ export async function resolveReader({
...
 	try {
 		const graph = await getProjectGraph(projectRootDir, ui5ConfigPath);
-		const project = graph.getRoot();
+		project = graph.getRoot();
 		projectRootDir = project.getRootPath();
 		fsBasePath = project.getSourcePath();
 		fsBasePathTest = path.join(projectRootDir, project._testPath ?? "test");
 
-		if (!namespace && !project._isSourceNamespaced) {
+		if (!project._isSourceNamespaced) {
```
matz3, line 342: I wonder why this logic here needs to exist. There is already a mapping of the virtual file paths to the filesystem paths in place at https://github.com/SAP/ui5-linter/blob/3694911541e7da5f6581f91fa2504812055056b1/src/linter/linter.ts#L76 which seems to work fine.
I might of course miss some detail here, but I would expect that the very same logic is needed to rewrite the paths so that the fs based ignores / files can be matched.

d3xter666: Anyway, you're correct that this functionality is already there in the `lintProject`!
I'll try to figure out if it's possible to refactor it in a way that it also works for `lintFiles`

src/linter/linter.ts
```diff
@@ -114,11 +118,12 @@ export async function lintFile({
 }
 
 async function lint(
-	resourceReader: AbstractReader, options: LinterOptions, config: UI5LintConfigType
+	resourceReader: AbstractReader, options: LinterOptions & {relFsBasePath: string; virBasePath: string},
```
matz3, line 121: Why is the `& {relFsBasePath: string; virBasePath: string}` needed? Is it because those are mandatory here, but optional in the `LinterOptions`?

src/linter/LinterContext.ts
```diff
@@ -58,13 +58,18 @@ export interface TranspileResult {
...
-	namespace?: string;
 	filePatterns?: FilePattern[];
 	ignorePattern?: FilePattern[];
 	reportCoverage?: boolean;
 	includeMessageDetails?: boolean;
 	configPath?: string;
 	ui5ConfigPath?: string;
+	namespace?: string;
+
+	relFsBasePath?: string;
```
matz3, line 69: Is this really the right interface for those options? It is also used by `lintProject` and `lintFile`, but the options are not considered there, which is confusing.

matz3 (changes requested):
Please add tests for library projects to ensure the "filePaths" option also works there, as that was the main issue here.

src/linter/LinterContext.ts
d3xter666: Ah, nice catch! Thanks!

src/linter/linter.ts
d3xter666: Now it's refactored

test/fixtures/linter/projects/sap.ui.core/src/ui5loader.js
matz3, line 1: Do we really need the real ui5loader.js file here? I think linting e.g. the existing `sap/ui/base/Object.js` would cover the same case of linting a file outside of the library namespace, isn't it?

matz3: True, but previously the linting failed regardless of whether there were findings or not, so for that matter it is not relevant. But I'm also fine with the reduced ui5loader.js.

matz3 (approved):

PR #370, opened 2024-10-18
fix(amdTranspiler): Fix exception "Node not found in array"

matz3:
When multiple return statements are found in the AMD factory, the
factory function will not be replaced, so no assignment of the old
factory block should be done.
This caused the exception "Node not found in array" as replacements
were wrongly applied on the new top level SourceFile instead of the
original and still existing factory function.


d3xter666 (approved):
LGTM

codeworrior:
Would you mind explaining in the commit, what the problem was? 

I assume, the fact that the only return statement does not refer to the name (symbol) of the class?

matz3:
@codeworrior done. Thanks for asking. This made me questioning the code again, as some parts were not covered. Now it should work as expected in all cases.

codeworrior (approved):
Thanks for updating the commit message and extended fixtures! That helped me to better understand the issue behind.

PR #371, opened 2024-10-18
fix: Properly check UI5 class inheritance

matz3:
The UI5 class inheritance check needs to perform a different check when
the UI5 base classes are actually within the current project like it is
the case for the sap.ui.core library.
In this case the declaration is based on the actual code, not the UI5
type definitions.

This change adds a very minimal sap.ui.core project where this scenario
can be reproduced without having to test the whole original library.

This is a follow-up of https://github.com/SAP/ui5-linter/pull/349


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -152,8 +152,20 @@ export default class SourceFileLinter {
 					return parentClassType.symbol?.declarations?.flatMap((declaration) => {
```
codeworrior, line 152: off topic: what is the `flatMap` needed for? Why was `some` not sufficient?

codeworrior (approved):
LGTM

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Can't recall the exact reasoning, but after look at it, definitelly `some` would work perfectly fine here

matz3: I kept the `flatMap` during my refactoring as I wasn't sure whether it might be helpful in case we want to collect information about the inheritance. But right now I don't think that it is necessary, so using `some` would make sense to me.

d3xter666: https://github.com/SAP/ui5-linter/pull/366/commits/37f9ff27e90dbea8144e386442f0239493d926b1

PR #374, opened 2024-10-18
feat: Check control renderer declaration

matz3:
Checks the renderer declaration of a UI5 Control and reports an error if deprecated usage is detected.

JIRA: CPOUI5FOUNDATION-876


src/linter/ui5Types/amdTranspiler/rewriteExtendCall.ts
```diff
@@ -66,42 +66,79 @@ function getClassBodyFromArguments(
...
 			// Use method declarations as-is
 			// e.g. "method() {}"
+
+			// Special handling:
+			// - renderer: *static*
+			// This aligns it with how UI5 projects should declare those properties in TypeScript
+
+			if (
+				ts.isIdentifier(prop.name) && prop.name.text === "renderer" &&
+				!prop.modifiers?.find((mod) => mod.kind === ts.SyntaxKind.StaticKeyword)
```
d3xter666, line 76: `.some`?
```suggestion
				!prop.modifiers?.some((mod) => mod.kind === ts.SyntaxKind.StaticKeyword)
```

src/linter/messages.ts
```diff
@@ -457,4 +460,30 @@ export const MESSAGE_INFO = {
...
+
+	[MESSAGE.CONTROL_RENDERER_DECLARATION_STRING]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-deprecated-control-renderer-declaration"],
+
+		message: ({className, rendererName}: {className: string; rendererName: string | undefined}) =>
+			`Deprecated declaration of renderer ${rendererName ? `'${rendererName}' ` : ""}for Control '${className}'`,
+		details: ({className, rendererName}: {className: string; rendererName: string | undefined}) => {
+			const rendererModuleName = rendererName ? `'${rendererName.replace(/\./g, "/")}'` : "renderer";
+			return `Defining the 'renderer' for the control '${className}' as string may lead to synchronous loading ` +
```
codeworrior, line 483: Suggested edit: "...as string..." --> "..by its name..."

d3xter666 (dismissed):

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -124,81 +124,118 @@ export default class SourceFileLinter {
...
 				}
 
 				ts.forEachChild(childNode, visitMetadataNodes);
 			};
-
-			if (this.isUi5controlMetadataNode(node)) {
-				ts.forEachChild(node, visitMetadataNodes);
-			}
+			ts.forEachChild(node, visitMetadataNodes);
+		} else if (ts.isClassDeclaration(node) && this.isUi5ClassDeclaration(node, "sap/ui/core/Control")) {
```
codeworrior, line 139: would it make sense to make the isClassDeclaration check part of the isUi5ClassDeclaration check?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -124,81 +124,118 @@ export default class SourceFileLinter {
...
 										declaration.parent.parent &&
 										ts.isModuleDeclaration(declaration.parent.parent) &&
-										declaration.parent.parent.name?.text === "sap/ui/base/ManagedObject"
+										declaration.parent.parent.name?.text === baseClass.module
 									) ||
 									// Declaration via real class (within sap.ui.core project)
 									(
 										ts.isSourceFile(declaration.parent) &&
-										declaration.parent.fileName === "/resources/sap/ui/base/ManagedObject.js"
+										declaration.parent.fileName === `/resources/${baseClass.module}.js`
```
codeworrior, line 175: Does currently not happen for sap.ui.core code, but what if the code would be in TypeScript in future? (*.ts)
Could happen if we ever introduce a check against a sap.fe base class?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -124,81 +124,118 @@ export default class SourceFileLinter {
...
+			// No definition of renderer causes the runtime to load the corresponding Renderer module synchronously
+			this.#reporter.addMessage(MESSAGE.MISSING_CONTROL_RENDERER_DECLARATION, {className}, node);
+			return;
 		}
 
-		if (!parentNode) {
-			metadataFound();
-			return false;
-		}
+		if (ts.isPropertyDeclaration(rendererMember) && rendererMember.initializer) {
```
codeworrior, line 216: Somehow my last comment went lost, strange.

I wondered why the filter above accepts MethodDeclarations but the check here does not care about them.
Is that because the check for apiVersion:2 is only part of the follow-up BLI?

matz3: First I had the check but then removed the handling of MethodDeclarations as they are not considered deprecated. I missed to adjust the filter, as it actually wasn't on prupose.

But you're right. When we check for the renderer itself, it probably needs to be added in there again, so I would just leave it there.

src/linter/ui5Types/SourceFileLinter.ts
matz3: Good catch. I will also add a `.ts` extension check.

src/linter/messages.ts
```diff
@@ -480,8 +480,8 @@ export const MESSAGE_INFO = {
 			`Deprecated declaration of renderer ${rendererName ? `'${rendererName}' ` : ""}for Control '${className}'`,
 		details: ({className, rendererName}: {className: string; rendererName: string | undefined}) => {
 			const rendererModuleName = rendererName ? `'${rendererName.replace(/\./g, "/")}'` : "renderer";
-			return `Defining the 'renderer' for the control '${className}' as string may lead to synchronous loading ` +
-				`of the ${rendererModuleName} module. ` +
+			return `Defining the 'renderer' for control '${className}' as by its name may lead to synchronous ` +
```
codeworrior, line 483: My English isn't good enough to fully judge this, but was "as by..." intended?

matz3: Nope 🙈
Thanks 👍🏻

codeworrior (approved):
LGTM (just a small language question)

codeworrior (dismissed):
LGTM

src/linter/messages.ts
```diff
@@ -457,4 +460,30 @@ export const MESSAGE_INFO = {
...
+			`If no renderer exists, set 'renderer: null'. Otherwise, either import the renderer module ` +
+			`and assign it to the 'renderer' property or implement the renderer inline.`,
+	},
+
+	[MESSAGE.CONTROL_RENDERER_DECLARATION_STRING]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-deprecated-control-renderer-declaration"],
+
+		message: ({className, rendererName}: {className: string; rendererName: string | undefined}) =>
+			`Deprecated declaration of renderer ${rendererName ? `'${rendererName}' ` : ""}for Control '${className}'`,
```
KlattG, line 480:
```suggestion
			`Deprecated declaration of renderer ${rendererName ? `'${rendererName}' ` : ""}for control '${className}'`,
```

KlattG (changes requested):
Just one small thing.

PR #376, opened 2024-10-21
fix(amdTranspiler): Handle extend call without args

matz3:
This fixes two exceptions that occurred when a call to an "extend"
function without arguments was encountered:
- Cannot read properties of undefined (reading 'getSourceFile')
- Cannot read properties of undefined (reading 'text')

The calls are now handled properly and will be ignored as their
signature does not match the expected one for UI5 classes.


codeworrior (approved):
LGTM.

PR #392, opened 2024-10-31
fix: Improve Control Renderer lint

d3xter666:
- Detect non static renderer objects in ES6/TS code
- Detect apiVersion property that is not part of the variable initialization
- Properly analyze SAPUI5 raw renderers that might be ambient modules and present in `@sapui5/types`


src/linter/messages.ts
```diff
@@ -333,6 +334,15 @@ export const MESSAGE_INFO = {
 		details: () => `"{@link sap.ui.core.RenderManager#methods/icon RenderManager}",`,
 	},
 
+	[MESSAGE.NOT_STATIC_CONTROL_RENDERER]: {
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-deprecated-api"],
+
+		message: ({className}: {className?: string}) =>
+			`The control renderer ${className ? ("of '" + className + "'") : ""} must be a static object`,
```
codeworrior, line 342: I would rather say "...must be a static __property__". Otherwise, one might think that it must be an embedded renderer. But for sure, references are also okay.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -208,7 +208,6 @@ export default class SourceFileLinter {
 		const className = node.name?.getText() ?? "<unknown>";
 		const rendererMember = node.members.find((member) => {
 			return (ts.isPropertyDeclaration(member) || ts.isMethodDeclaration(member)) &&
-				member.modifiers?.some((modifier) => modifier.kind === ts.SyntaxKind.StaticKeyword) &&
```
codeworrior, line 211: My thinking was that we should only complain about a non-static `renderer` property when there is no static one. It should be allowed for controls to define a valid (static) renderer and use an instance property with the same name. Only when no static renderer is defined, an instance property IMO indicates a wrong pattern.

The `members.find()` will only find the first occurrence. I would therefore suggest to keep the old check and introduce a new check only in the case that there no `rendererMember` was found (around line 226 in the old code).

src/linter/messages.ts
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -56,14 +57,15 @@ export default class SourceFileLinter {
 
 	constructor(
 		context: LinterContext, resourcePath: ResourcePath,
```
codeworrior, line 59: Maybe [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) would come in handy for constructors like this one

src/linter/messages.ts
```diff
@@ -333,6 +334,15 @@ export const MESSAGE_INFO = {
 		details: () => `"{@link sap.ui.core.RenderManager#methods/icon RenderManager}",`,
 	},
 
+	[MESSAGE.NOT_STATIC_CONTROL_RENDERER]: {
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-deprecated-api"],
+
+		message: ({className}: {className?: string}) =>
+			`The control renderer ${className ? ("of '" + className + "'") : ""} must be a static property`,
```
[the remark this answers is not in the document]
d3xter666: Yep, that should be exactly it.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -223,6 +225,19 @@ export default class SourceFileLinter {
 			])) {
 				return;
 			}
+
+			const nonStaticRender = (node as ts.ClassDeclaration).members.find((member: ts.ClassElement) => {
+				return (ts.isPropertyDeclaration(member) || ts.isMethodDeclaration(member)) &&
+					(ts.isIdentifier(member.name) || ts.isStringLiteral(member.name)) &&
```
codeworrior, line 231: I wonder whether the check for equality works when the import was a relative import (`import MyRenderer from "./MyRenderer.js"`)

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -255,8 +270,54 @@ export default class SourceFileLinter {
...
+					// raw file in order to analyze the renderer. Such special case, is for example
+					// any openui5 library when we analyze the openui5 repo- we have the TS definitions
+					// loaded in the TS compiler via @sapui5/types, but we actually want to analyze the
+					// source files of those types.
+					let importModuleString = "";
+
+					// Get the import string module from the current source file, so we can filter later
+					const importedRenderModule = this.#sourceFile.statements
+						.find((statement) => (ts.isImportDeclaration(statement) &&
+							statement.importClause?.name?.getText() === rendererMember.initializer?.getText()));
```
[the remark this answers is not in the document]

src/linter/ui5Types/SourceFileLinter.ts
Very convenient, indeed!

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Is this test case what you meant: https://github.com/SAP/ui5-linter/pull/392/commits/976992b348a09d5ca4de9a85f0acd42827523475 ?

If, yes, then yes, it works even with relative imports

codeworrior: Not sure if this proves that the check works as expected.

To my understanding, the issue in `7ControlRenderer` is already found because of `7Control`. So what is the indication that the import of `7ControlRenderer` by `Test5` also leads to an analysis of `7ControlRenderer`? 

Shouldn't this rather be a new renderer (which is not reached otherwise) and if it appears in the snapshot, then it's working?

Well, but I might have missed something.

d3xter666: This check is done, but not reported as it is a duplicate entry and we clean duplicates.
I'll create a separate renderer file. Thanks!

d3xter666: https://github.com/SAP/ui5-linter/pull/392/commits/102b7b1e5904a17298d3d23f3fbc05f7e9454b74

src/linter/messages.ts
```diff
@@ -333,6 +334,15 @@ export const MESSAGE_INFO = {
 		details: () => `"{@link sap.ui.core.RenderManager#methods/icon RenderManager}",`,
 	},
 
+	[MESSAGE.NOT_STATIC_CONTROL_RENDERER]: {
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-deprecated-api"],
```
matz3, line 339: I wouldn't expect this an `deprecated API`, as it never was intended to be defined that way. Therefore I think we should use a different ruleId. Also, it is not related to UI5 2.x, but a general check.

src/linter/messages.ts
matz3, line 342: There will be two spaces in case `className` is not provided.

src/linter/messages.ts
```diff
@@ -333,6 +334,15 @@ export const MESSAGE_INFO = {
...
+
+		message: ({className}: {className?: string}) =>
+			`The control renderer ${className ? ("of '" + className + "'") : ""} must be a static property`,
+		details: () => "",
```
matz3, line 343: Rather return `undefined` when there are no details available.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -236,13 +224,26 @@ export default class SourceFileLinter {
 			])) {
 				return;
 			}
+
+			const nonStaticRender = (node as ts.ClassDeclaration).members.find((member: ts.ClassElement) => {
+				return (ts.isPropertyDeclaration(member) || ts.isMethodDeclaration(member)) &&
+					(ts.isIdentifier(member.name) || ts.isStringLiteral(member.name)) &&
+					member.name.text === "renderer";
+			});
+			if (nonStaticRender) {
```
matz3, line 233: Shouldn't this check be done before the check for the special cases above:
```ts
this.isUi5ClassDeclaration(node, [
				"sap/ui/core/mvc/View",
				// XMLComposite is deprecated, but there still shouldn't be a false-positive about a missing renderer
				"sap/ui/core/XMLComposite",
				"sap/ui/core/webc/WebComponent",
				"sap/uxap/BlockBase",
			])
```
?

Otherwise a non static renderer of those subclasses won't be detected.

src/linter/messages.ts
d3xter666: Done

src/linter/messages.ts
d3xter666: True, thanks!
But, to me this one seems too specific. Would `static-var-required` be ok, then?

src/linter/messages.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Makes sense to move the before the spacial cases as it would never be able to detect that otherwise.

src/linter/messages.ts
matz3: Maybe `ui5-class-declaration`, as it is about how UI5 classes are declared/defined.
What do you think @codeworrior / @flovogt / @RandomByte ?

flovogt: just another idea `ui5-control-definition`?

codeworrior: As the `renderer` is not a general concept for UI5 classes, I rather tend to `ui5-control-declaration`. But there might not be much in addition to this check that would justify a separate rule-ID for that. So I would be fine with `ui5-class-declaration`, too.

d3xter666: Changed it to `ui5-class-declaration`. Let me know if something else comes

matz3 (approved):

PR #394, opened 2024-10-31
fix: Check renderer declarations in sources only

RandomByte:


matz3:
This is obsolete now and handled via #392, right?

PR #400, opened 2024-11-06
feat: Node API in ui5 linter

d3xter666:
JIRA: CPOUI5FOUNDATION-885


package.json
```diff
@@ -33,12 +33,12 @@
...
 		"cleanup": "rimraf lib coverage",
 		"coverage": "nyc ava --node-arguments=\"--experimental-loader=@istanbuljs/esm-loader-hook\"",
-		"depcheck": "depcheck --ignores @commitlint/config-conventional,@istanbuljs/esm-loader-hook,rimraf,sap,mycomp",
+		"depcheck": "depcheck --ignores @commitlint/config-conventional,@istanbuljs/esm-loader-hook,rimraf,sap,mycomp,@ui5/linter",
 		"hooks:pre-push": "npm run lint:commit",
 		"lint": "eslint .",
 		"lint:commit": "commitlint -e",
 		"prepare": "node ./.husky/skip.js || husky",
-		"test": "npm run lint && npm run build-test && npm run coverage && npm run depcheck",
+		"test": "npm run lint && npm run build && npm run coverage && npm run depcheck",
```
d3xter666, line 41: We now need a real build, in order to test the `package.json` exports

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       run: npm run depcheck
 
     - name: Perform build
-      run: npm run build-test
+      run: npm run build
```
d3xter666, line 34: We now need a real build, in order to test the package.json exports

test/fixtures/package-exports.ts
```diff
@@ -0,0 +1,22 @@
...
+});
+
+// Check number of defined exports
+test.serial("Package exports: check number of exports", (t) => {
+	const packageJson = require("@ui5/linter/package.json");
+	t.is(Object.keys(packageJson.exports).length, 2);
+});
+
+// Public API contract (exported modules)
+test.serial("Package exports: @ui5/linter", async (t) => {
```
matz3, line 18: This is the only test which requires a TS build. What about moving it away from the unit tests into an `e2e` test (e.g. at `test/e2e/` similar to what we have already in ui5-cli. The test can still be an ava test, but shouldn't be included into the default set that runs via `npm run unit`. This way it would not create overhead for all other tests.

package.json
```diff
@@ -26,20 +26,21 @@
...
-		"depcheck": "depcheck --ignores @commitlint/config-conventional,@istanbuljs/esm-loader-hook,rimraf,sap,mycomp",
+		"depcheck": "depcheck --ignores @commitlint/config-conventional,@istanbuljs/esm-loader-hook,rimraf,sap,mycomp,@ui5/linter",
 		"hooks:pre-push": "npm run lint:commit",
 		"lint": "eslint .",
 		"lint:commit": "commitlint -e",
 		"prepare": "node ./.husky/skip.js || husky",
-		"test": "npm run lint && npm run build-test && npm run coverage && npm run depcheck",
-		"unit": "ava",
+		"test": "npm run lint && npm run build && npm run coverage && npm run depcheck",
+		"unit": "npm run build && ava",
```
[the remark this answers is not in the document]
d3xter666: Basically, `test/fixtures/package-exports.ts` requires the index to be build.

ava.config.js
```diff
@@ -10,6 +10,7 @@ export default {
 	},
 	files: [
 		"test/lib/**/*.ts",
+		"test/fixtures/package-exports.ts",
```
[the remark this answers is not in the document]
d3xter666: The issue is that `fixtures` are excluded from typescript config. 
The import of `@ui5/linter` in the `/lib` is ambiguous to TypeScript as it creates circular refference and TS cannot detremine which is the root folder.

src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+	ui5Config = "./ui5.yaml",
+	rootDir = process.cwd(),
+}: UI5LintOptions): Promise<LintResult[]> {
+	return lintProject({
+		rootDir,
+		filePatterns,
+		ignorePattern: ignorePatterns,
+		reportCoverage: coverage,
+		includeMessageDetails: details,
+		configPath: config,
```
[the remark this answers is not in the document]
d3xter666: I did rename with some of them, which made sense, but `configPath` to me seems more explanatory.
I'll take another look at the options

ava-e2e.config.js
```diff
@@ -0,0 +1,6 @@
+// Calculate nodeArguments based on the Node version
```
matz3, line 1: This comment is obsolete, also in the original file `ava.config.js`

eslint.config.js
```diff
@@ -19,6 +19,7 @@ export default tseslint.config(
 			"test/tmp/*",
 			"test/projects/*",
 			"test/fixtures/*",
+			"test/e2e/*",
```
matz3, line 22: Why shouldn't we lint the e2e tests? Only test fixtures and tmp files should be excluded

tsconfig.json
```diff
@@ -14,5 +14,6 @@
 		"test/tmp/**/*",
 		"test/projects/**/*",
 		"test/fixtures/**/*",
+		"test/e2e/**/*",
```
matz3, line 17: Also here, I would expect that e2e tests are handled the same as our existing tests at `test/lib/`. They should be built so that TypeScript errors can be reported and fail our CI.

ava-e2e.config.js
d3xter666: Done

eslint.config.js
d3xter666: Because of the imports of itself `test/e2e/package-exports.ts` must be excluded from `eslint` and `typescript` as they got confused by it. I'll exclude just this file, then.

tsconfig.json
d3xter666: Because of the imports of itself `test/e2e/package-exports.ts` must be excluded from `eslint` and `typescript` as they got confused by it. I'll exclude just this file, then.

d3xter666: If tests for `test/e2e/package-exports.ts` fail, the CI will be failed

src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+// in the options and behavior compared to LinterOptions internal type.
+export interface UI5LintOptions {
+	filePatterns?: string[];
+	ignorePatterns?: string[];
+	details?: boolean;
+	config?: string;
+	noConfig?: boolean;
+	coverage?: boolean; // boolean(default: false)
+	ui5Config?: string | object;
+	rootDir?: string;
```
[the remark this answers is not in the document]
d3xter666: Done

src/index.ts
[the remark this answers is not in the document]
d3xter666: Done

eslint.config.js
d3xter666: I have added some comments in the code why we need to exclude this file in eslint

tsconfig.json
d3xter666: I have added some comments in the code why we need to exclude this file in tsconfig

src/index.ts
```diff
@@ -0,0 +1,69 @@
+import {lintProject} from "./linter/linter.js";
+import type {LintResult} from "./linter/LinterContext.js";
+
+export type {LintResult} from "./linter/LinterContext.js";
+export type {UI5LintConfigType} from "./utils/ConfigManager.js";
+
+// Define a separate interface for the Node API as there could be some differences
+// in the options and behavior compared to LinterOptions internal type.
+export interface UI5LinerOptions {
```
matz3, line 9:
```suggestion
export interface UI5LinterOptions {
```

test/lib/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	const results = await Promise.all([
+		ui5lint({rootDir: appPath}),
+		ui5lint({rootDir: libPath}),
+	]);
+
+	results.forEach((res) => {
+		t.snapshot(preprocessLintResultsForSnapshot(res));
+	});
+});
```
matz3, line 137: `ui5lint()` should be callable without any arguments to get the all defaults. Currently this fails.

src/index.ts
```diff
@@ -0,0 +1,69 @@
+import {lintProject} from "./linter/linter.js";
+import type {LintResult} from "./linter/LinterContext.js";
+
+export type {LintResult} from "./linter/LinterContext.js";
+export type {UI5LintConfigType} from "./utils/ConfigManager.js";
```
matz3, line 5: is this type relevant / necessary for outside consumers?

src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+	details?: boolean;
+	/**
+	 * Path to a ui5lint.config.(cjs|mjs|js) file
+	 */
+	config?: string;
+	/**
+	 * Whether to skip loading of the ui5lint.config.(cjs|mjs|js) config file
+	 * @default false
+	 */
+	noConfig?: boolean;
```
matz3, line 31: did we discuss / decide on this option vs. `config?: string | boolean` to suppress the default config?

src/index.ts
d3xter666: Thanks!

src/index.ts
d3xter666: Thanks! I have forgotten to remove it

src/index.ts
d3xter666: Nope, but here are my 2 cents on that:
As we are trying to be aligned with ESLint, there it's a [`--no-config-lookup`](https://eslint.org/docs/latest/use/command-line-interface#options) option which does excatly the same thing.

For the sake of discussion, here are the options we have listed:
1. boolean | string (default / true: "ui5lint.config.{js,mjs,cjs}"; false should suppress reading the default config file)
2. string (default: "ui5lint.config.{js,mjs,cjs}"; empty string should suppress reading the default config file)
3. separate noConfig option to suppress reading the default config file

Let's see and stick with one of these options.

test/lib/index.ts
d3xter666: Done

test/lib/index.ts
```diff
@@ -121,6 +121,18 @@ test.serial("ui5lint API: Ignore config file for com.ui5.troublesome.app", async
 	t.snapshot(preprocessLintResultsForSnapshot(res));
 });
 
+test.serial("ui5lint API: Use defaults", async (t) => {
+	const projectPath = path.join(fixturesProjectsPath, "com.ui5.troublesome.app");
+	const {ui5lint, sinon} = t.context;
+	// Stub process.cwd(), so we have some app to test
+	sinon.stub(process, "cwd").returns(projectPath);
+
+	const res = await ui5lint({});
```
matz3, line 130: Please also add another test case where no arguments are provided, as this is what previously failed.

src/index.ts
RandomByte: Actually, we discussed (and apparently agreed on) yet another option in CPOUI5FOUNDATION-885 😅 

> - **config: (null | string (default: "ui5lint.config.{js,mjs,cjs}"; null should suppress reading the default config file, a string overwrites the default path)**
> 
>     - _Alternatives discussed (decision above):_
>        - boolean | string (default / true: "ui5lint.config.{js,mjs,cjs}"; false should suppress reading the default config file)
>        - string (default: "ui5lint.config.{js,mjs,cjs}"; empty string should suppress reading the default config file)
>        - separate noConfig option to suppress reading the default config file

If there's a need, we can of course discuss it again tomorrow!

test/lib/index.ts
d3xter666: Thanks!
Sorry, I have forgotten to remove that

src/index.ts
d3xter666: As discussed today, we'll go with the current approach- a `noConfig` oprions

matz3 (commented):
LGTM

PR #406, opened 2024-11-12
fix: Allow ignoring ui5.yaml files via config

matz3:
For some files, only the passed ignorePatterns were considered, but not pattern defined in the configuration.

This mainly affected linting of ui5.yaml files, as they are accessed via the so-called root reader rather than the project workspace, where the problem did not occur.


RandomByte (commented):
LGTM

src/linter/linter.ts
```diff
@@ -163,7 +163,7 @@ async function lint(
 		reader,
 	});
 
-	const res = await lintWorkspace(workspace, filePathsWorkspace, options, config, matchedPatterns);
+	const res = await lintWorkspace(workspace, filePathsWorkspace, options, config, ignorePattern, matchedPatterns);
```
d3xter666, line 166: The whole build ignorePattern does not need to be provided in lintWorkspace.
The configuration is also available there. The same has already been done for the filePatterns there: https://github.com/SAP/ui5-linter/blob/06871b660db5ed01799abff8558cd4f8f59dc835/src/linter/lintWorkspace.ts#L27

src/linter/lintWorkspace.ts
```diff
@@ -33,7 +36,7 @@ export default async function lintWorkspace(
 		patternsMatch,
 	});
 	reader = resolveReader({
-		patterns: options.ignorePattern ?? [],
+		patterns: ignorePattern ?? [],
```
d3xter666, line 39: All the information is available already, so something like this would work:

[
		...(config.ignores ?? []),
		...(options.ignorePattern ?? []),
	].filter(($) => $);

matz3: True, that also works. I though of not duplicating the merging logic. I'll refactor it.

matz3: Done

d3xter666 (approved):

Issue #373, opened 2024-10-18
Ignore check of unmatched patterns


Shingaz:
## Is your feature request related to a problem? Please describe.

It is very difficult to use the ui5linter in conjuction with [lint-staged](https://github.com/lint-staged/lint-staged). To avoid to lint the whole project before commiting, we use lint-staged lint only files staged for commit.

The lint-staged configuration is this one:
```js
const ui5lint = (filenames) => `ui5lint ${filenames.join(" ")}`;

export default {
  "webapp/**/*": (stagedFiles) => ui5lint(stagedFiles),
};
```

Depending on what is staged for commit, some files can be included in the resulting command but ui5lint is not able to process them. For example:
```sh
ui5lint webapp/test/integration/mock/metadata.xml webapp/view/Main.view.xml 
```

In this case, as `webapp/test/integration/mock/metadata.xml` cannot be processed by ui5lint, the whole command fails.

## Describe the solution you'd like

Have a parameter to ignore files that cannot be processed. Similarly to Prettier [`--ignore-unknown`](https://prettier.io/docs/en/cli.html#--ignore-unknown) for example.

## Describe alternatives you've considered

Monkey patch the ui5-linter library.

flovogt:
@Shingaz Thanks a lot for opening this enhancement request. Great Idea! We will implement this feature with backlog item CPOUI5FOUNDATION-955

PR #408, opened 2024-11-13
fix: Detect skipped & unsupported files [PoC]

d3xter666:
fixes: https://github.com/SAP/ui5-linter/issues/373

### DIFF
--- a/src/linter/linter.ts
+++ b/src/linter/linter.ts
@@ -169,7 +169,7 @@ async function lint(
 	});
 
 	const res = await lintWorkspace(workspace, filePathsWorkspace, options, config, matchedPatterns);
-	checkUnmatchedPatterns(filePatterns, matchedPatterns);
+	await checkUnmatchedPatterns(filePatterns, matchedPatterns, options);
 
 	lintEnd();
 	return res;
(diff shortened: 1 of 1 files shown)


src/linter/linter.ts
```diff
@@ -372,6 +372,33 @@ function checkUnmatchedPatterns(patterns: FilePattern[], patternsMatch: Set<stri
 		return acc;
 	}, [] as FilePattern[]);
 
+	// Check if the Glob pattern is actually a file that exists, but the linter is not able to process.
+	if (unmatchedPatterns.length) {
+		const notProcessedFiles = new Set<string>();
+		const rootFileReader = createReader({
+			fsBasePath: options.rootDir,
+			virBasePath: "/",
```
RandomByte, line 380: We can exclude node_modules, .git etc by providing an `excludes` parameter here:

```suggestion
			virBasePath: "/",
			excludes: [""]
```

src/linter/linter.ts
```diff
@@ -372,6 +372,33 @@ function checkUnmatchedPatterns(patterns: FilePattern[], patternsMatch: Set<stri
...
+		});
+		// TODO: Define a better way to get all files. Exclude node_modules, .git, etc.
+		const allFiles = await rootFileReader.byGlob("/**/*");
+		const filePaths = allFiles.map((file) => file.getPath());
+
+		for (const pattern of unmatchedPatterns) {
+			if (filePaths.some((filePath) => filePath.includes(pattern))) {
```
RandomByte, line 387: `pattern` could also be something like `"**/*.pdf"`

d3xter666: Thanks!
Now we check files against the actual pattern

PR #409, opened 2024-11-14
fix(Test Starter Configs): Fix false-positive findings

maxreichmann:
This fixes a false-positive finding of the property `theme` inside [Configuration.beforeBootstrap.qunit.js](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/test/sap/ui/core/qunit/bootstrap/Configuration.beforeBootstrap.qunit.js). This resulted in undefined runtime errors.
`theme` should only be checked for Test Starter configs. 

Changes:
* Now, the **file name** is checked against a Regex and optional chaining was optimized to prevent these runtime errors further.

* A sample `Configuration.beforeBootstrap.qunit.js` was added as a test artifact. As seen in the snapshots, there are no more findings for this file.

* In the course of adding a Regex check for the file name, `testsuite2.qunit.js` was renamed to `testsuite.special-case-quotes.qunit.js` to be aligned with framework functionality.


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1308,21 +1308,25 @@ export default class SourceFileLinter {
 	}
 
 	analyzeTestsuiteThemeProperty(node: ts.PropertyAssignment) {
+		// Check if the node is part of a testsuite config file by its file name.
+		// This is the same check as in the framework and prevents false-positives.
+		const validTestSuiteName = /^\/testsuite(?:\.[a-z][a-z0-9-]*)*\.qunit\.(?:js|ts)$/;
```
matz3, line 1313: Please mention the original location of this pattern (_utils in openui5).

test/fixtures/linter/rules/NoDeprecatedApi/Configuration.beforeBootstrap.qunit.js
```diff
@@ -0,0 +1,10 @@
+// This legacy file was a False Positive for the new Test Starter detections.
```
matz3, line 1: It's not a legacy file. It is part of the Configuration test in the core library. Maybe add a reference to the original location in openui5, or we come up with a minimal example where the same problem can be reproduced.

matz3 (approved):

maxreichmann:
https://github.com/SAP/ui5-linter/commit/cf166f7a10c9513dd184447eec8f3145a7d5d8a0 was removed due to commit message violations and replaced with https://github.com/SAP/ui5-linter/commit/cf6907c54eef58d5a68dcd3d5cb3ad15b2f7fe35.

PR #423, opened 2024-11-22
feat: Add type support for "byId" in controllers

matz3:
This feature adds type definitions for `byId` within controllers so that
the usage of deprecated functionality can be detected when controls or
elements from the view are accessed by their ID.
Previously, only the base element class was known, which is still the
case as a fallback.

JIRA: CPOUI5FOUNDATION-843


src/linter/xmlTemplate/generator/ControllerByIdDtsGenerator.ts
```diff
@@ -0,0 +1,65 @@
...
+	constructor(private controllerByIdInfo: ControllerByIdInfo) {
+	}
+
+	generate() {
+		const mappings = this.controllerByIdInfo.getMappings();
+		if (mappings.size === 0) {
+			return null;
+		}
+		let out = "";
+		this.controllerByIdInfo.getMappings().forEach((idToModules, controllerName) => {
```
d3xter666, line 16:
```suggestion
		mappings.forEach((idToModules, controllerName) => {
```

matz3: Thanks, I missed to update that 👍🏻

RandomByte (approved):
LGTM

src/linter/xmlTemplate/generator/ControllerByIdDtsGenerator.ts
```diff
@@ -0,0 +1,75 @@
...
+		out += "\t}\n";
+		return out;
+	}
+
+	private generateModuleDeclaration(controllerName: string, idToModules: IdModulesMap) {
+		const moduleName = controllerName.replace(/\./g, "/") + ".controller";
+
+		let out = `declare module "${moduleName}" {\n`;
+		out += this.generateByIdMapping(idToModules);
+		out += `\ttype ByIdFunction = {\n`;
```
RandomByte, line 47: Template string might be more readable but this is fine too.

matz3: Thanks. I also didn't like that part and I think in-between refactorings I had reasons for not using a template literal. I took the chance to also go over the other parts with focus on improving readability.

RandomByte (approved):

PR #430, opened 2024-11-28
feat: Support projects with type "module"

matz3:
This allows linting projects that currently do not fit any other project
structure of UI5 Tooling.

To lint such projects, a `ui5.yaml` needs to be added to the root of the
project with the following content:

```yaml
specVersion: "4.0"
type: module
metadata:
  name: <project-name>
resources:
  configuration:
    paths:
      /resources/<namespace>/: "<sources-folder>"
````

Placeholders `<project-name>`, `<namespace>` and `<sources-folder>` need
to be replaced with the actual values.

Example:
A project has its sources in a folder `src` and uses the namespace
`my.project`, so for example a source file named
`src/my/project/util/Formatter.js` would exist.
In this case the placeholders would be replaced as follows:
- `<project-name>`: `my.project`
- `<namespace>`: `my/project`
- `<sources-folder>`: `src`

### DIFF
--- a/src/linter/linter.ts
+++ b/src/linter/linter.ts
@@ -6,7 +6,7 @@ import {taskStart} from "../utils/perf.js";
 import path from "node:path";
 import posixPath from "node:path/posix";
 import {stat} from "node:fs/promises";
-import {ProjectGraph} from "@ui5/project";
+import {Project, ProjectGraph} from "@ui5/project";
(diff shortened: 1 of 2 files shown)


matz3:
Open tasks
- [ ] Tests
- [ ] Update README

matz3:
Also see https://github.com/SAP/ui5-linter/pull/148

PR #437, opened 2024-12-03
fix: Detect deprecations in new expressions with ID

matz3:
This solves the gap of not detecting usage of deprecated UI5 metadata
(e.g. properties, aggregations, events) when using the constructor
with an ID as first argument.


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -681,37 +681,41 @@ export default class SourceFileLinter {
 		}
 
 		// There can be multiple and we need to find the right one
-		const [constructSignature] = classType.getConstructSignatures();
+		const allConstructSignatures = classType.getConstructSignatures();
```
RandomByte, line 684: If `!node.arguments.length` we can skip most of this, right?

RandomByte (changes requested):

src/linter/ui5Types/SourceFileLinter.ts
matz3: Done

RandomByte (approved):

PR #440, opened 2024-12-06
fix: Improve module resolution

matz3:
This solves issues where some detections did not work properly due to
the way dependencies were resolved.

When using relative imports, the module resolution was working, but when
for example importing a control from a test, the complete namespace is
used, but the corresponding JS file was not found.

This change ensures that JS files are resolved and also that the types
for a framework library are loaded when it is linted.


d3xter666 (commented):
LGTM, but it seems coverage needs to be either improved or decreased a bit

matz3:
I checked the difference and in `SourceFileLinter.ts` there is now less code covered in `analyzeControlRendererDeclaration`, as it takes a different path for the fixed case.

@d3xter666 do you know whether there is still a case that needs the handling in here?
https://github.com/SAP/ui5-linter/blob/e9a5812fec5bc3262174508ffce66d67a801bf9c/src/linter/ui5Types/SourceFileLinter.ts#L315-L331

d3xter666:
> I checked the difference and in `SourceFileLinter.ts` there is now less code covered in `analyzeControlRendererDeclaration`, as it takes a different path for the fixed case.
> 
> @d3xter666 do you know whether there is still a case that needs the handling in here?
> 
> https://github.com/SAP/ui5-linter/blob/e9a5812fec5bc3262174508ffce66d67a801bf9c/src/linter/ui5Types/SourceFileLinter.ts#L315-L331

It seems that this test/file [`test/fixtures/linter/projects/sap.f/src/sap/f/ProductSwitch.js`](https://github.com/SAP/ui5-linter/blob/main/test/fixtures/linter/projects/sap.f/src/sap/f/ProductSwitch.js)
 does not go through the `moduleExportsSymbols...` loop anymore. It's resolved here:
https://github.com/SAP/ui5-linter/blob/main/src/linter/ui5Types/SourceFileLinter.ts#L286

I believe this is caused by the new namespace resolve logic in here and also the conditional inclusion of the `.d.ts` files... I'm starting to question whether we need this custom module resolution logic anymore. But I suppose there might be cases where we intentionally might include the `.d.ts` files and then, we'll definitely need that

matz3:
@d3xter666 thanks a lot 👍🏻 Then I'll go ahead with that. We can still revert it in case we encounter some relevant cases again.

d3xter666 (approved):

PR #443, opened 2024-12-11
fix: Allow renderer v4

d3xter666:
Currently the linter check only if renderer is explicitly defined with version 2.
There's also version 4 for it. We need to take into account this, so to avoid false positives


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -346,13 +346,16 @@ export default class SourceFileLinter {
 		if (node && (ts.isObjectLiteralExpression(node) || ts.isVariableDeclaration(node))) {
 			const apiVersionNode = findApiVersionNode(node) as ts.PropertyAssignment | ts.NumericLiteral | undefined;
 
+			const availableApiVersions = ["2", "4"];
```
RandomByte, line 349: Maybe there'll be a v5 some day which we probably shouldn't report as deprecated?

codeworrior: I think we initially limited this to "2" to be able to complain about "unknown" api versions. The same is now true for "4". 
But I get your point how to keep this up-to-date with the runtime capabilities.

d3xter666: Yep, I was wondering for that, too.
An easy solution would be `< 2`, but we still don't have 5 and also there isn't (officially) version 3.
So, setting those values would produce false positives. To be honest, I'm not sure what's the renderer algorithm that detects versioning... would it accept values other than `2` and `4` and if yes, what would be the result?

codeworrior: Runtime fact: The runtime defaults a missing apiVersion to 1 and contains checks whether the `version === 1` (or `!== 1`)  and another check whether `apiVersion === 4`.

That being set, any `version > = 1` would be fine technically, **BUT** documentation explicitly defines contracts for versions 1, 2 and 4 only, not for any other version. So no other version should be specified by control developers.

codeworrior: As my reply might not be clear enough: I'm in favour of the current check for 2 or 4 as valid, everything else is not supported in UI5 2.0 (as of now).

RandomByte (approved):

PR #446, opened 2024-12-13
feat(linter): Extend project initialization fallbacks

RandomByte:
Enhance compatibility for legacy UI5 projects that do not include a
ui5.yaml and do not use current best practice directory structures like
'webapp' for apps and 'src'/'test' for libraries.

In addition to those directories we now also check and eventually use
the following directories:

Applications:
* src/main/webapp
* WebContent

Libraries:
* src/main/jslib (+ src/test/jslib)
* src/main/uilib (+ src/test/uilib)
* src/main/js (+ src/test/js)

### DIFF
--- a/src/linter/linter.ts
+++ b/src/linter/linter.ts
@@ -188,26 +188,38 @@ async function getProjectGraph(rootDir: string, ui5Config?: string | object): Pr
 		rootConfigPath = ui5YamlPath;
 	} else {
 		if (ui5Config) throw new Error(`Unable to find UI5 config file '${ui5Config}'`);
-		const isApp = await dirExists(path.join(rootDir, "webapp"));
-		if (isApp) {
-			rootConfiguration = {
-				specVersion: "3.0",
-				type: "application",
-				metadata: {
-					name: "ui5-linter-target",
-				},
-			};
-		} else {
-			const isLibrary = await dirExists(path.join(rootDir, "src"));
-			if (isLibrary) {
-				rootConfiguration = {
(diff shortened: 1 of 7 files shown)


src/linter/linter.ts
```diff
@@ -188,26 +188,24 @@ async function getProjectGraph(rootDir: string, ui5Config?: string | object): Pr
...
+		if (await dirExists(path.join(rootDir, "webapp"))) {
+			// Common app with webapp folder
+			rootConfiguration = createProjectConfig("application", "webapp");
+		} else if (await dirExists(path.join(rootDir, "src", "main", "webapp"))) {
+			// Legacy app with src/main/webapp folder
+			rootConfiguration = createProjectConfig("application", "src/main/webapp");
+		} else if (await dirExists(path.join(rootDir, "WebContent"))) {
+			// Legacy app with WebContent folder
+			rootConfiguration = createProjectConfig("application", "WebContent");
+		} else if (await dirExists(path.join(rootDir, "src", "main", "jslib"))) {
```
codeworrior, line 200: In the "big set", I found one project with src/main/js/ (AP-FIORI-CORE/nw.core.ushell_abap), it's not correctly handled by the logic that I've shared.

src/main/jslib, src/main/uilib don't occur in the only occur in the "big set", but in SAPUI5 contributor libs

For WebContent, I couldn't find an example, I guess this structures used by very old tools and might occur in BSP repos only, not on GitHub.

d3xter666 (dismissed):
LGTM

PR #452, opened 2024-12-19
refactor(update-sapui5-types script): Adjust metadata generation of UI5 types

maxreichmann:
JIRA: CPOUI5FOUNDATION-973

### 1. Restructure of format in `api-extract.json`:
- [x] Generate metadata of `aggregation`, `defaultAggregation`, `association` , `event`, `property` values (incl. borrowed ones for UI5Classes)
  **Note**: options `method` and `constructor` are not getting generated as not being required

This metadata structure was decided upon:
```
{...
  "sap.ca.ui.charts.Chart": {
        "properties": [
          "width",
          "height",
          "dataset",
          "chartType",
          "minShapeSize"
        ],
        "events": [...],
        "aggregations": [...],
        "associations": [...],
        "defaultAggregation": "content",
        "extends": "sap.ui.core.Control" // Placeholder which will only be generated by Approach B
   },
   ...
}
```

- [x] Add toggle switch for generating metadata with two approaches (see 2.)
- [x] Provide methods for reading this new structure of metadata
- [x] Update existing collection of `defaultAggregation` values
- [x] Check existing usage of `api-extract.json` (linting + tests)
- [x] Add script for measuring performances of two approaches (see 2.)
- [x] Perform measurements and document
---

### 2. Comparison of performance/filesize impacts for two approaches:
**Approach A:** Write entire metadata (incl. borrowed options) to `api-extract.json`.

Filesize: ~1.7MB
Performance:
image

**Approach B:** Write only metadata which is not borrowed by parent objects by using a placeholder "extends". During runtime, read `api-extract.json` and collect all metadata by resolving "extends".

Filesize: ~0.8MB
Performance:
image

##### Test settings:

- SAPUI5 types version used: **1.120.25**
- command used for tests:
```sh
npx tsx scripts/_testPerformance/api-extractJSON_reader.ts
```
- number of runs for each test: **1000**
- To switch between the two approaches, use the boolean toggle in https://github.com/SAP/ui5-linter/blob/bd90322616fbf6c6caf17cdd90acaba7a513c4f2/scripts/metadataProvider/createMetadataInfo.ts#L59 and run again `scripts/update-sapui5-types.ts`

---

### 3. Conclusion:
As expected, Approach B results in a significantly smaller file size of `api-extract.json` than Approach A.
In addition, Approach B performed similar during runtime in all test cases, although extra resolving of the placeholders is necessary.
For this reason, my recommendation is to implement Approach B in the ui5-linter.

### 4. Implementation:
Approach B was decided upon.


matz3:
I like the current structure and that information is not repeated. In addition we could now also include the deprecation texts (see `ApiExtract.getDeprecationInfo`) into the same structure, right?

maxreichmann:
> I like the current structure and that information is not repeated. In addition we could now also include the deprecation texts (see `ApiExtract.getDeprecationInfo`) into the same structure, right?

The full JSON structure looks like this now:

```
{
  framework: {
	  name: "SAPUI5",
	  version: sapui5Version,
  },
  metadata: {
          // all UI5 symbol records...
  },
  deprecations: {
	  UI5Class: {},
	  UI5Enum: {},
	  UI5Function: {},
	  UI5Namespace: {},
	  UI5Typedef: {},
	  UI5Interface: {},
  },
};
```

So deprecations are in the same place as before and also have the same structure as before.
`ApiExtract.getDeprecationInfo` still works (I've added some unit tests for this method).

matz3:
Yeah, my idea was to move them into `metadata`, like this:

```json
{
  "sap.ca.ui.charts.Chart": {
    "properties": [],
    "events": [],
    "aggregations": [],
    "associations": [],
    "defaultAggregation": "content",
    "extends": "sap.ui.core.Control",
    "deprecation": "<deprecation-text>"
  }
}
```

scripts/metadataProvider/MetadataProvider.ts
```diff
@@ -45,4 +62,104 @@ export default class MetadataProvider {
...
+				}
+				break;
+			/* case "UI5Enum": // Disabled, as UI5Enum does not contain any option values (with 1.131.1)
+				symbolMetadata = symbol as UI5Enum;
+				break;
+			case "UI5Function": // Disabled, as UI5Function does not contain any option values (with 1.131.1)
+				symbolMetadata = symbol as UI5Function;
+				break; */
+			case "UI5Namespace":
+				symbolMetadata = symbol as UI5Namespace;
```
matz3, line 130: Is it necessary that we still return an empty object instead of undefined here (and in the other types below)?
If the empty object entry in the JSON is not require, we should rather skip it to reduce the file size.

src/utils/ApiExtract.ts
```diff
@@ -17,19 +32,98 @@ interface DeprecationInfo {
...
+	 * @example getAllOptionsByType("sap.m.App", "properties", false)
+	 * 	returns ["backgroundColor", "backgroundImage", "backgroundRepeat", *etc.*]
+	 * @example getAllOptionsByType("sap.ui.base.Object", "properties")
+	 * 	returns []
+	 * @example getAllOptionsByType("sap.m.App", "notExistingOption")
+	 * 	returns undefined
+	 * @example getAllOptionsByType("sap.xyz.NotExistingSymbol", "properties")
+	 * 	returns undefined
+	 */
+	getAllOptionsByType(symbolName: string, optionType: AllowedMetadataOptionType, resolveBorrowed = false):
```
matz3, line 62: Here it's called `resolveBorrowed`, but in the interface it's `includeBorrowed`. I would slightly prefer `includeBorrowed` because it more explicitly describes what the result is.

src/utils/ApiExtract.ts
```diff
@@ -17,19 +32,98 @@ interface DeprecationInfo {
...
+	 * Returns the type of a given option in a UI5 symbol record.
+	 * (resolving borrowed ones is optional)
+	 * @example getTypeByOption("sap.m.App", "backgroundColor", false)
+	 * 	returns "properties"
+	 * @example getTypeByOption("sap.m.App", "notExistingOption")
+	 * 	returns undefined
+	 * @example getTypeByOption("sap.xyz.notExistingSymbol", "backgroundColor")
+	 * 	returns undefined
+	 */
+	getTypeByOption(symbolName: string, optionName: string, resolveBorrowed = false):
```
matz3, line 93: same here

test/lib/utils/ApiExtract.ts
```diff
@@ -0,0 +1,78 @@
...
+		"UI5Class": {
+			"sap.f.Avatar": "(since 1.73) Use the {@link sap.m.Avatar} instead."
+		}
+	}
+}`;
+
+const apiExtract = new ApiExtractImpl(JSON.parse(mockApiJSON));
+
+test("Test getAllOptionsByType()", (t) => {
+	t.deepEqual(apiExtract.getAllOptionsByType("sap.m.App", "properties", false),
```
matz3, line 32: The default behaviour of the optional last argument should also be tested.

test/lib/utils/ApiExtract.ts
```diff
@@ -0,0 +1,78 @@
...
+	t.deepEqual(apiExtract.getAllOptionsByType("sap.ui.base.Object", "properties", false), [],
+		"If no option values could be found, an empty array should be returned");
+	t.is(apiExtract.getAllOptionsByType("sap.xyz.notExistingSymbol", "properties", false), undefined,
+		"If the symbol can't be found, undefined should be returned");
+	t.is(apiExtract.getAllOptionsByType("sap.xyz.notExistingSymbol", "properties", true), undefined,
+		"If the symbol can't be found, undefined should be returned");
+});
+
+test("Test getTypeByOption()", (t) => {
+	t.is(apiExtract.getTypeByOption("sap.m.App", "backgroundColor", false), "properties",
```
matz3, line 49: The default behaviour of the optional last argument should also be tested.

scripts/metadataProvider/MetadataProvider.ts
maxreichmann: It definitely would make sense to return undefined here and not write to the file to save space.
But this would apply for all symbol kinds (e.g. "UI5Class", "UI5TypeDef", ..) and not only "UI5Namespace", right?

matz3: Yes, it applies to all of them.

test/lib/utils/ApiExtract.ts
maxreichmann: Done

src/utils/ApiExtract.ts
maxreichmann: Done

src/utils/ApiExtract.ts
maxreichmann: Done

scripts/metadataProvider/MetadataProvider.ts
maxreichmann: Done

matz3 (commented):
LGTM, apart from the performance test files and the obsolete "usePlaceholder" flag in MetadataProvider.ts

matz3 (approved):

matz3:
> > Yeah, my idea was to move them into `metadata`, like this:
> > ```json
> > {
> >   "sap.ca.ui.charts.Chart": {
> >     "properties": [],
> >     "events": [],
> >     "aggregations": [],
> >     "associations": [],
> >     "defaultAggregation": "content",
> >     "extends": "sap.ui.core.Control",
> >     "deprecation": "<deprecation-text>"
> >   }
> > }
> > ```
> 
> As decided, this will be tackled in a follow-up PR.

It turned out the this actually increased the file size, so we'll leave it as it is for now.

Issue #458, opened 2025-01-02
crashes with error ERR_UNSUPPORTED_ESM_URL_SCHEME on Windows with cwd on D:


dinurp:
## Expected Behavior

Runs

## Current Behavior

Crashes

## Steps to Reproduce the Issue

On a windows machine execute `npx --yes @ui5/linter@latest xx.view.xml --verbose`

## Context

- UI5 linter version: **`1.5.0`**
- Node.js version: **`v22.12.0`**
- npm version: `unknown`
- OS/Platform: `Windows`
- Browser *(if relevant)*: `unknown`
- Other information regarding your environment *(optional)*: `working directory on d:`

## Log Output / Stack Trace

```
PS D:\temp\zz> npx --yes @ui5/linter@latest xx.view.xml --verbose
verb cli:middlewares:base using ui5lint version 1.5.0 (from C:\Users\xx\AppData\Local\npm-cache\_npx\6e4ede960b3b258c\node_modules\@ui5\linter\bin\ui5lint.js)
verb cli:middlewares:base using node version v22.12.0

⚠️  Process Failed With Error

Error Message:
Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'

Stack Trace:
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:217:11)
    at defaultLoad (node:internal/modules/esm/load:109:3)
    at ModuleLoader.load (node:internal/modules/esm/loader:670:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:483:43)
    at #createModuleJob (node:internal/modules/esm/loader:507:36)
    at #getJobFromResolveResult (node:internal/modules/esm/loader:275:34)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:243:41)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:546:25)

If you think this is an issue of the ui5-linter, you might report it using the following URL: https://github.com/SAP/ui5-linter/issues/new/choose
```

RandomByte:
[...]

PR #460, opened 2025-01-03
fix(ConfigManager): Resolve ESM import exception on Windows

RandomByte:
`import(path)` does not accept absolute Windows-paths since those seem to get confused with URLs.

Before, this was mitigated by always calculating a relative path (relative to the UI5 linter install directory). This however breaks on Windows if the target project is located on another drive as reported in https://github.com/SAP/ui5-linter/issues/458

Instead convert all paths to `file://`-URLs before passing them to `import()`.

Also, in linter.ts, enforce the project root path to always be absolute to simplify handling the path later on.

Resolves https://github.com/SAP/ui5-linter/issues/458


d3xter666:
LGTM

matz3 (approved):

PR #474, opened 2025-01-10
fix(parsing-error): Failed to map back to source

matz3:
For files without code but with comments it could happen that a
parsing-error is caused when trying to map back a finding to the
sources.

This is now prevented by not taking a source map into account when it
does not contain any mappings. Checking whether a TraceMap is empty does
not seem to be possible with public API.


d3xter666 (approved):
LGTM

RandomByte (approved):

PR #475, opened 2025-01-10
feat: Detect deprecated 'type' in `View.create` / `<mvc:View>`

matz3:
This adds the detection of the partial deprecation of the View.create factory.

In addition, nested views within XML that use the general <mvc:View> tag are now also detected when using a deprecated 'type'.

JIRA: CPOUI5FOUNDATION-907


d3xter666 (commented):
LGTM

RandomByte (approved):

PR #478, opened 2025-01-14
feat: Detect deprecated 'type' in `Fragment.load` / `<core:Fragment>`

matz3:
This adds the detection of the partial deprecation of the Fragment.load
factory.

In addition, nested fragments within XML are now also detected when
using a deprecated 'type'.

JIRA: CPOUI5FOUNDATION-907


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1163,6 +1165,42 @@ export default class SourceFileLinter {
...
+
+		if (typeProperty && ts.isStringLiteralLike(typeProperty.initializer)) {
+			const typeValue = typeProperty.initializer.text;
+			if (typeValue === "HTML") {
+				this.#reporter.addMessage(MESSAGE.PARTIALLY_DEPRECATED_FRAGMENT_LOAD, {
+					typeValue,
+				}, node);
+			}
+		}
+	}
```
d3xter666, line 1202: This code is 99% like the `#analyzeViewCreate`. 
Is it possible to consolidate those and just do the check & message separately

d3xter666: Maybe, we need to somehow extract and reuse the type field extraction. I have it also in my change:
https://github.com/SAP/ui5-linter/pull/480/files#diff-7bc0071c1214325807d3a82a222b097360e3b0020aae68d0575dbae4aa43b99aR1151

d3xter666 (commented):
LGTM

d3xter666 (approved):

src/linter/ui5Types/SourceFileLinter.ts
matz3: Yeah, some better reuse would be great.

PR #480, opened 2025-01-14
feat: Detection for UI5 Model Data Types

d3xter666:
JIRA: CPOUI5FOUNDATION-914


test/lib/linter/rules/snapshots/NoDeprecatedApi.ts.md
```diff
@@ -1908,6 +2102,14 @@ Generated by [AVA](https://avajs.dev).
             ruleId: 'no-globals',
             severity: 2,
           },
+          {
+            column: 3,
```
matz3, line 2106: This is a duplicate finding. The first one is created during XML analysis (but seems to have a wrong line/column info) and the new one comes from the transformed JS code.

d3xter666: Thanks! Now the xml views & fragments are skipped in this check. They are being analyzed in the XML Parser

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1262,6 +1262,18 @@ export default class SourceFileLinter {
 					prop.initializer.text.startsWith("{") && prop.initializer.text.endsWith("}") &&
 					ts.isNewExpression(node) &&
 					this.#isPropertyBindingType(node, prop.name.getText())) {
+					const nodeSourceMap = this.sourceMaps?.get(this.resourcePath);
+					const traceMap = nodeSourceMap ? new TraceMap(nodeSourceMap) : undefined;
```
matz3, line 1266: Doing this check in here might be expensive, as the source map needs to be parsed over and over again.
What about instead storing the required information during xml transpiling in the `LintMetadata` and then just accessing it in here to decide whether a check should be performed or not?

d3xter666: We need the traceMap for the original position of the node in the XML view.

matz3: Yes, I understand that. But not in case we actually want to skip our check in JS, as the same check has already been performed during XML transpilation. So I wonder in which case we still need to report findings of XML sources in SourceFileLinter. If we do need it, we should only parse the sourceMap in that case, not to just check for the original file name. I suspect this would cause a lot of unnecessary overhead during linter runtime.

d3xter666: Done

src/linter/xmlTemplate/linter.ts
```diff
@@ -37,6 +37,11 @@ export default async function lintXml({filePathsWorkspace, workspace, context}:
 		await workspace.write(transpiledResource);
 		await filePathsWorkspace.write(transpiledResourceSourceMap);
 		await workspace.write(transpiledResourceSourceMap);
+
+		// Stash information that this .js file is actually a transpiled XML.
+		const contextMeta = context.getMetadata(jsPath);
+		contextMeta.xmlCompiledResources = contextMeta.xmlCompiledResources ?? new Set();
```
matz3, line 43: Why storing a Set instead of just a flag that the resource has been transformed from XML (e.g. boolean)?
Isn't there always a 1:1 relation from XML source file to JS file?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1230,120 @@ export default class SourceFileLinter {
...
+				if (typeField && ts.isPropertyAssignment(typeField) &&
+					(!ts.isNewExpression(node) /* bindProperty() */ ||
+						this.#isPropertyBindingType(node, prop.name.getText()) /* new Control() */)) {
+					this.#analyzeModelTypeField(typeField.initializer);
+				} else if (ts.isStringLiteral(prop.initializer) &&
+					prop.initializer.text.startsWith("{") && prop.initializer.text.endsWith("}") &&
+					ts.isNewExpression(node) &&
+					this.#isPropertyBindingType(node, prop.name.getText())) {
+					const originalFilename = Array.from(
+						this.context.getMetadata(this.resourcePath)?.xmlCompiledResources ?? [])[0];
```
matz3, line 1266: metadata is already read within the `lint` method, so it could be stored as an instance member there and then accessed here.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1230,120 @@ export default class SourceFileLinter {
...
+					let resourcePath = this.resourcePath;
+					if (originalFilename) {
+						if ([".view.xml", ".fragment.xml"].some((ending) => originalFilename.endsWith(ending))) {
+							// Do not process xml-s. This case would be handled separately withing the BindingParser
```
matz3, line 1270: What about doing this check at the start of this method to ignore all NewExpression nodes in that case. I don't see why any of the above checks are necessary to be performed.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1230,120 @@ export default class SourceFileLinter {
 		}
 	}
 
+	#analyzeModelDataTypes(node: ts.NewExpression | ts.CallExpression) {
```
matz3, line 1233: Handling both cases, `bindProperty` and constructors within one method and especially one loop over the arguments sounds overly complex and error prone to me.

Imagine a control has a property called "type" (string). With this current implementation, it would be falsely detected as global type usage, in case a normal string is passed in a constructor call.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1230,120 @@ export default class SourceFileLinter {
...
+						});
+
+					const nodeSourceMap = this.sourceMaps?.get(this.resourcePath);
+					const traceMap = nodeSourceMap ? new TraceMap(nodeSourceMap) : undefined;
+					const {start: nodePos} = getPositionsForNode({
+						node: prop,
+						sourceFile: this.sourceFile,
+						resourcePath: this.resourcePath,
+						traceMap: traceMap,
+					});
```
matz3, line 1293: As we won't check code that is transpiled from XML, I wouldn't expect that this special handling is needed at all.

src/linter/xmlTemplate/linter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1231,116 @@ export default class SourceFileLinter {
 		}
 	}
 
+	#analyzeModelDataTypes(node: ts.NewExpression | ts.CallExpression) {
+		const originalFilename = this.#metadata?.xmlCompiledResource;
+		// Do not process xml-s. This case would be handled separately withing the BindingParser
+		if (originalFilename && [".view.xml", ".fragment.xml"].some((ending) => originalFilename.endsWith(ending))) {
```
matz3, line 1237: I wonder in which (future) case, we would not have an `*.xml` filename stored in the `xmlCompiledResource` metadata. Isn't a boolean sufficient here?

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks! Indeed, that's the case!

However, the interface of the `bindProperty()` and `new Control()` is pretty much the same with the difference that in control there might be more levels of nesting.
I have now added a guard against it and also enhanced the tests.

matz3: I still find it not easy to understand that both cases are handled in the same loop over the arguments.
Especially, as there is already an iteration over the `node.arguments` of a `NewExpression` (the one that checks for deprecations). Wouldn't it be natural to include the binding check in there?

d3xter666: Those find different things. The new expression with `ObjectLiteralExpression` has the same analysis as the `bindProperty` as they both have `ObjectLiteralExpression` as an argument:
- new expression: https://github.com/SAP/ui5-linter/pull/480/files#diff-122df669e89998d5fadc1eb7a7dd0ba69b93887a901b75d12371612f81f7947aR2
- bindProperty: https://github.com/SAP/ui5-linter/pull/480/files#diff-445fdeb2d93b7c2c6fb00d828adff3e521f5b882cb2345a7b5d1425ae7d7c634R7

So, when we find the `type` property in those object literals, we can check whether the `type` is a module that's already loaded.

The case with a string binding is completely different- we have a property with a special "string" inside: https://github.com/SAP/ui5-linter/pull/480/files#diff-7e46da9d18fd803a3726945a53a496331ee866a6f73351bf6638dded02c12172R3
So, the `value` in this case is a special, binding string literal which is directly passed to the BindingParser. 
The difference is that `value`'s `.initializer` is now string literal that needs special handling, but not a `ObjectLiteralExpression` where we must search for the `type` field. It's not possible here.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: But then we'll need to resolve the filename. I don't mind either, but this one seems very convenient, at least for this case

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: I have refactored the code, so that `bindProperty()`, `bind*()` and `new Control()` analysis receive the same structure- the eventual binding `ObjectLiteralExpression`

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -682,6 +685,14 @@ export default class SourceFileLinter {
 			this.#analyzeNewCoreRouter(node);
 		} else if (moduleDeclaration?.name.text === "sap/ui/model/odata/v4/ODataModel") {
 			this.#analyzeNewOdataModelV4(node);
+		} else if (nodeType.symbol.declarations?.some(
+			(declaration) => this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control"))) {
+			const originalFilename = this.#metadata?.xmlCompiledResource;
+			// Do not process xml-s. This case would be handled separately withing the BindingParser
```
maxreichmann, line 691:
```suggestion
			// Do not process xml-s. This case would be handled separately within the BindingParser
```

maxreichmann:
Maybe move all fixture tests from `test/fixtures/linter/rules/NoDeprecatedApi` to `test/fixtures/linter/rules/NoGlobals` to match the ruleId?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1245,120 @@ export default class SourceFileLinter {
...
+	 * Identify if a constructor argument property is bindable
+	 *
+	 * Get the class declaration by going up the chain and then find the argument property
+	 * from there. Directly finding the type of the property is not possible from here as it's
+	 * missing some context.
+	*/
+	#isPropertyBindingType(node: ts.NewExpression, propName: string) {
+		const controlAmbientModule =
+			this.getSymbolModuleDeclaration(this.checker.getTypeAtLocation(node).symbol);
+		const classArg = (controlAmbientModule?.body as ts.ModuleBlock)?.statements
```
matz3, line 1326: Isn't body defined as `ts.ModuleBody`? Please refrain from using `as` unless there is really no other solution. This poses the risk of bugs and errors in the code as the following code might rely on the wrong type.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1245,120 @@ export default class SourceFileLinter {
...
+			?.find((stmnt): stmnt is ts.ClassDeclaration => stmnt.kind === ts.SyntaxKind.ClassDeclaration)
+			?.members.find((m): m is ts.ConstructorDeclaration => ts.isConstructorDeclaration(m));
+
+		const constructorArgType = classArg && this.checker.getTypeAtLocation(classArg.parameters[0]);
+		const argProperty = constructorArgType?.getProperties()
+			.find((p: ts.Symbol) => p.name === propName);
+
+		const argPropType = argProperty?.declarations?.[0] &&
```
matz3, line 1334: This exact declaration of a property within a NewExpression is already available in the existing iteration over the constructor arguments in `analyzeNewExpression`.

The `argPropType` can be accessed there via `this.checker.getTypeAtLocation(propertySymbol.valueDeclaration)`.

test/lib/linter/xmlTemplate/snapshots/transpiler.ts.md
```diff
@@ -578,13 +578,25 @@ Generated by [AVA](https://avajs.dev).
       {
         column: 3,
         line: 9,
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1245,120 @@ export default class SourceFileLinter {
...
+
+			// Get the type property
+			let typeField;
+			if (ts.isObjectLiteralExpression(prop.initializer)) {
+				typeField = prop.initializer.properties.find((prop: ts.ObjectLiteralElementLike) => {
+					return ts.isPropertyAssignment(prop) &&
+						ts.isIdentifier(prop.name) &&
+						prop.name.text === "type";
+				});
+			} else if (ts.isStringLiteral(prop.initializer) &&
```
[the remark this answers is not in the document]
d3xter666: Replaced all the occurences in the file

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1223,6 +1245,120 @@ export default class SourceFileLinter {
 		}
 	}
 
+	#analyzeNewAndApplySettings(node: ts.NewExpression | ts.CallExpression) {
+		node?.arguments?.filter((arg) => ts.isObjectLiteralExpression(arg))
+			.flatMap((arg) => arg.properties)
+			.forEach((prop) => {
+				if (ts.isPropertyAssignment(prop) &&
+					(ts.isCallExpression(node) ||
+						this.#isPropertyBindingType(node, prop.name.getText()))
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -274,7 +274,7 @@ export default class SourceFileLinter {
 			if (initializerType.flags & ts.TypeFlags.StringLiteral) {
 				let rendererName;
 				if (
-					ts.isStringLiteral(rendererMember.initializer) ||
+					ts.isStringLiteralLike(rendererMember.initializer) ||
 					ts.isNoSubstitutionTemplateLiteral(rendererMember.initializer)
```
matz3, line 278: This check is obsolete now as `StringLiteralLike` includes `NoSubstitutionTemplateLiteral`.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1059,7 +1081,7 @@ export default class SourceFileLinter {
 	}
 
 	#analyzeThemingSetTheme(node: ts.CallExpression) {
-		if (!node.arguments.length || !ts.isStringLiteral(node.arguments[0])) {
+		if (!node.arguments.length || !ts.isStringLiteralLike(node.arguments[0])) {
```
matz3, line 1084: Ah, I've missed that in my refactoring.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1166,120 @@ export default class SourceFileLinter {
...
+				});
+			} else if (ts.isStringLiteralLike(prop.initializer) &&
+				ts.isIdentifier(prop.name) && prop.name.text === "type" &&
+				// Whether it's a direct property, named "type" of the Control
+				// or type in property binding
+				!ts.isNewExpression(prop.parent.parent)) {
+				typeField = prop;
+			}
+
+			if (typeField && ts.isPropertyAssignment(typeField)) {
```
matz3, line 1208: Can typeField be something else than a PropertyAssignment, in case it has a value?

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1166,120 @@ export default class SourceFileLinter {
...
+			}
+
+			// Get the type property
+			let typeField;
+			if (ts.isObjectLiteralExpression(prop.initializer)) {
+				typeField = prop.initializer.properties.find((prop: ts.ObjectLiteralElementLike) => {
+					return ts.isPropertyAssignment(prop) &&
+						ts.isIdentifier(prop.name) &&
+						prop.name.text === "type";
+				});
```
matz3, line 1199: Can be replaced with `getPropertyAssignmentInObjectLiteralExpression` from the utils.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: TS resolves it as `ts.ObjectLiteralElementLike | undefined`, so we need this check

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -830,6 +842,16 @@ export default class SourceFileLinter {
 				!VALID_TESTSUITE.test(this.sourceFile.fileName) &&
 				symbolName === "ready" && moduleName === "sap/ui/core/Core") {
 				this.#reportTestStarter(node);
+			} else if (symbolName === "applySettings" &&
+				nodeType.symbol.declarations?.some((declaration) =>
+					this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control"))) {
+				this.#analyzeNewAndApplySettings(node);
+			} else if (symbolName.startsWith("bindProperty") && moduleName === "sap/ui/base/ManagedObject") {
+				this.#analyzeModelDataTypes(node.arguments[1] as ts.ObjectLiteralExpression);
```
matz3, line 850: Please refrain from using `as`. This here skips the check for undefined values, so undefined might get passed, but the function "analyzeModelDataTypes" doesn't expect undefined to be passed, so it misses checks for undefined values, causing a runtime exception.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Thanks!

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: With the refactoring, now it seem to be working

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1166,123 @@ export default class SourceFileLinter {
...
+			if (!ts.isPropertyAssignment(prop)) {
+				return;
+			}
+
+			// Get the type property
+			let typeField;
+			if (ts.isObjectLiteralExpression(prop.initializer)) {
+				typeField = propNames
+					.map((name) => getPropertyAssignmentInObjectLiteralExpression(
+						name, prop.initializer as ts.ObjectLiteralExpression))
```
d3xter666, line 1197: This casting is inevitable as TS does not detect properly that `prop.initializer` is `ObjectLiteralExpression` althought the if statement above

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1166,123 @@ export default class SourceFileLinter {
...
+					if (ts.isObjectLiteralExpression(prop.initializer)) {
+						this.#analyzeControlBindings(prop.initializer, ["type"]);
+					} else {
+						this.#analyzeControlStringBindings(prop);
+					}
+				}
+			});
+	}
+
+	#analyzeControlBindings(node: ts.ObjectLiteralExpression, propNames: string[]) {
```
matz3, line 1186: isn't this rather a PropertyBinding that is being analyzed here? AFAIK there's no control binding. Bindings are implemented on ManagedObject level, so if we need a more general name, it shouldn't be control, as also elements can have bindings but are not controls.

src/linter/ui5Types/SourceFileLinter.ts
matz3: This is because TypeScript can't be sure that between the `isObjectLiteralExpression` call and the usage within the map callback (different function context), the reference `prop.initializer` is still the same.

You can solve this by assigning `prop.initializer` to a local variable before the `.map` call, as in this case the reference can't change.

But maybe before changing that, please have a look at `getPropertyAssignmentsInObjectLiteralExpression` (plural form). Although I don't fully understand why the paramNames need to be passed from the outside, I think that would be the best approach to pick multiple properties from the given ObjectLiteralExpression.

src/linter/ui5Types/SourceFileLinter.ts
matz3: I actually wanted to motivate the re-use of that code so that no additional loop over the arguments and properties is needed at this place. But this can of course also be improved later on.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1166,121 @@ export default class SourceFileLinter {
...
+						moduleName: (importNode.moduleSpecifier as ts.StringLiteral).text,
+						variableName: importNode.importClause?.name?.text,
+					} as RequireDeclaration;
+				});
+
+			const {start: nodePos} = getPositionsForNode({
+				node: node,
+				sourceFile: this.sourceFile,
+				resourcePath: this.resourcePath,
+			});
```
matz3, line 1228: traceMap is missing here, so the line/column information might be wrong.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: I'm not quite sure how to reuse it. This implementation relies on ambient modules and their definitions.
Trying directly to find the type of an eventual propertyBinding does not provide enough context, so the union types and aliases cannot be determined correctly.

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
d3xter666: Done

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -830,6 +839,19 @@ export default class SourceFileLinter {
...
+			} else if (symbolName === "applySettings" &&
+				nodeType.symbol.declarations?.some((declaration) =>
+					this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control"))) {
+				this.#analyzeNewAndApplySettings(node);
+			} else if (["bindProperty", "bindAggregation"].includes(symbolName) &&
+				moduleName === "sap/ui/base/ManagedObject" &&
+				node.arguments[1] && ts.isObjectLiteralExpression(node.arguments[1])) {
+				this.#analyzePropertyBindings(node.arguments[1], ["type"]);
+			} else if (symbolName.startsWith("bind") &&
+				nodeType.symbol.declarations?.some((declaration) =>
```
matz3, line 851: `nodeType.symbol` might be undefined here. I'm not sure why the types do not express that.

Adding a fixture with `ControlEvents.bindAnyEvent({});` (imported from "sap/ui/events/ControlEvents") shows this issue.

Also, maybe an exact match for "bind" should be ignored here to not falsely analyze the native Function#bind method. But I'm not sure whether this will cause any issues here right now.

maxreichmann: `/bind[A-Z]+/.exec(symbolName)` ?

maxreichmann: `/bind[A-Z][a-z]*/.exec(symbolName)` ?

matz3: Properties and aggregations may contain uppercase letters (e.g. camelCase). Only the first letter of the method is always converted to uppercase, but after that it may be lowercase and uppercase and is not even limited to alphanumeric characters. The runtime seems to allow any valid JS property name, including unicode characters. And therefore properties could even start with a unicode character.

I think a robust check would rather be based on the types and to check the signature or argument types based on the signature (like it's already done for the NewExpression).

d3xter666: Detecting just the name seems quite fragile to me. The current implementation relies on:
- name starts with `bind`
- `bind*()` is called from an instance of `sap/ui/core/Control`

The only miss I see in this approach is if someone defines `bindSomething()` method within their control.
In such case, if we want to be completelly sure it's a binding, we must check method's signature, if its argument is of `PropertyBindingInfo` type

d3xter666: Now `bind*()` method is checked for `PropertyBindingInfo`.

maxreichmann (commented):
Just some minor changes. Apart from this, it looks very good 👍

Please include a fixture for the call of `bindAggregation()` of a managed object.

d3xter666:
> Just some minor changes. Apart from this, it looks very good 👍
> 
> Please include a fixture for the call of `bindAggregation()` of a managed object.

We have agreed to skip Aggregations form this change. This BLI only addresses `type` field which is missing in AggregationBinding, so adding it here gives no value

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1144,6 +1167,122 @@ export default class SourceFileLinter {
...
+
+		const constructorArgType = classArg && this.checker.getTypeAtLocation(classArg.parameters[0]);
+		const argProperty = constructorArgType?.getProperties()
+			.find((p: ts.Symbol) => p.name.toLocaleLowerCase() === propName);
+		const argPropType = argProperty?.valueDeclaration &&
+			this.checker.getTypeAtLocation(argProperty.valueDeclaration);
+
+		return argPropType?.isUnion() && argPropType.types.some((t) => {
+			return t?.getSymbol()?.getName() === "PropertyBindingInfo" ||
+				t.aliasSymbol?.getName() === "PropertyBindingInfo";
```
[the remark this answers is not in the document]
d3xter666: We have agreed to skip Aggregations form this change

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,216 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+    [
+      {
+        coverageInfo: [],
+        errorCount: 1,
+        fatalErrorCount: 0,
+        filePath: 'ModelDataTypes.view.xml',
+        messages: [
+          {
+            column: 9,
+            line: 2,
```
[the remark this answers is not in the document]
d3xter666: It's just a test sample here. This onw should be addressed in the BindingParser where XML validation happens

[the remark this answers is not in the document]
maxreichmann: So we should address this in another change.

maxreichmann (approved):

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -830,6 +839,20 @@ export default class SourceFileLinter {
 				!VALID_TESTSUITE.test(this.sourceFile.fileName) &&
 				symbolName === "ready" && moduleName === "sap/ui/core/Core") {
 				this.#reportTestStarter(node);
+			} else if (symbolName === "applySettings" &&
+				nodeType.symbol?.declarations?.some((declaration) =>
+					this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control"))) {
```
matz3, line 844: All related methods and functionality is implemented on the ManagedObject level, so we should always check for sub-classes of ManagedObject, not Control.

d3xter666: Fixed with: https://github.com/SAP/ui5-linter/pull/513

PR #482, opened 2025-01-14
refactor(metadata): Improve generating and reading

maxreichmann:
* refactor(metadata gen): Improve metadata identifier names
* refactor(metadata read): Improve provided methods
* test(ApiExtract): Replace mock data with real api-extract.json
* refactor(metadata gen): Detect exception for UI5 aggregations

### DIFF
--- a/src/utils/ApiExtract.ts
+++ b/src/utils/ApiExtract.ts
@@ -1,19 +1,18 @@
-/* eslint-disable @typescript-eslint/no-redundant-type-constituents */
 import {readFile} from "node:fs/promises";
 
 export type AllowedSymbolKind = "UI5Class" | "UI5Enum" | "UI5Interface" | "UI5Namespace" | "UI5Typedef" | "UI5Function";
-export type AllowedMetadataOptionType = "aggregations" | "associations" | "defaultAggregation" | "events" |
-	"methods" | "properties";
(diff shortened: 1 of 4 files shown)


RandomByte (approved):

src/utils/ApiExtract.ts
```diff
@@ -108,6 +113,66 @@ export class ApiExtractImpl implements ApiExtract {
...
 
+	/**
+	 * Checks whether a given aggregation is part of a UI5 symbol.
+	 * @example isAggregation("sap.m.NavContainer", "pages")
+	 * 	returns true
+	 * @example isAggregation("sap.m.App", "pages", false) // pages is a borrowed aggregation of "sap.m.NavContainer"
+	 * 	returns false
+	 */
+	isAggregation(symbolName: string, aggregationName: string, checkBorrowed = true): boolean {
+		return this.getAllOptionsByType(symbolName, "aggregation",
```
matz3, line 124: Currently, the check for a specific aggregation first collects ALL aggregations incl. borrowed and only then checks the list. This could be improved by already checking for the specific aggregation during the collection (currently the while-loop in `getTypeByOption `), which would reduce unnecessary collection and parent checks if the aggregation is already found.

This also applies to the other `isXY` methods.

matz3 (approved):

RandomByte (approved):

PR #489, opened 2025-01-20
fix: Improve overall code detection

matz3:
This change improves the overall code detection by refactoring the
SourceFileLinter class to make use of utility functions that increase
the coverage of AST node types that may appear and have not been checked
so far. This mainly includes TemplateLiterals by using
`isStringLiteralLike` instead of `isStringLiteral`.

In addition, the usage of `getText` has been replaced by actual node
checks and access to the actual properties (e.g. `text`). This makes
handling of quotes obsolete and is safer as whitespaces or line breaks
within the code might be included in the result of `getText`.


test/lib/linter/rules/snapshots/renderer.ts.md
```diff
@@ -43,6 +43,14 @@ Generated by [AVA](https://avajs.dev).
         ],
         warningCount: 0,
       },
+      {
+        coverageInfo: [],
+        errorCount: 0,
+        fatalErrorCount: 0,
+        filePath: '12Control.js',
```
d3xter666, line 50: Shouldn't this be reported?

matz3: Oh, I've missed that. Thanks for noticing 👍🏻

RandomByte (approved):
LGTM

PR #494, opened 2025-01-22
fix(XML): Do not fail on empty core:require attribute

matz3:
This change fixes the issue of causing an "parsing-error" when a
core:require attribute is empty. Although the attribute is obsolete and
should rather be removed, it is supported by the UI5 runtime, so linting
it should not cause a parsing error.

In addition, the types in the Parser have been improved and test
fixtures for error handling have been added.


codeworrior (approved):
LGTM

d3xter666 (approved):

PR #496, opened 2025-01-23
fix(binding): Resolve 'Bad name' error on tab

RandomByte:


RandomByte:
This PR is not the fix yet, it only adds fixtures that reproduce the error.

Note that OpenUI5's JSTokenizer does not show this behavior when inserting a tab after a property key. Also see Snippix No. 925080

codeworrior:
@RandomByte The JSTokenizer at runtime shows the same behaviour, but only when it parses a string with tabs originating from JS code. Or, in other words, it doesn't show the behaviour when parsing a string from XML. I don't find that in the spec, but I assume that the XML-Parser normalises whitespace in attribute values, e.g. converting from tab to space. When I use the following in your snippet:
```js
new sap.m.Button({text:"{path	: 'invoice>Status', formatter: 'String'}"});
```
it also fails with a "Bad name" error.

test/lib/linter/xmlTemplate/lib/JSTokenizer.ts
```diff
@@ -14,6 +14,15 @@ test("JSTokenizer: Basic values", (t) => {
 	});
 });
 
+test("JSTokenizer: Tab after key", (t) => {
+	const res = JSTokenizer.parseJS(
+		`{path	: 'invoice>Status', formatter: 'ui5.walkthrough.model.formatter.statusText'}`);
```
codeworrior, line 19: would you mind using \t for clarity?

codeworrior (commented):
LGTM, only one minor comment...

test/lib/linter/xmlTemplate/lib/JSTokenizer.ts
matz3: Done. Good point. I also added another test case with a quoted key, which did already work before. Only the unquoted failed.

codeworrior (approved):
LGTM

PR #499, opened 2025-01-24
feat(Formatter): Detect globals and strings of formatters in bindings (JS/TS)

maxreichmann:
JIRA: CPOUI5FOUNDATION-1005


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1240,21 +1241,20 @@ export default class SourceFileLinter {
...
 				?.find((stmnt): stmnt is ts.ClassDeclaration => stmnt.kind === ts.SyntaxKind.ClassDeclaration)
-				?.members.find((m): m is ts.ConstructorDeclaration => ts.isConstructorDeclaration(m));
+				?.members.find((m): m is ts.ConstructorDeclaration => m.kind === ts.SyntaxKind.Constructor);
 		}
 
 		const constructorArgType = classArg && this.checker.getTypeAtLocation(classArg.parameters[0]);
 		const argProperty = constructorArgType?.getProperties()
-			.find((p: ts.Symbol) => p.name === propName);
-		// this.checker.getTypeAtLocation(node.parent).members.get(propName).valueDeclaration.locals.get("oBindingInfo")
+			.find((p: ts.Symbol) => p.name.toLocaleLowerCase() === propName);
```
codeworrior, line 1257: Converting the property name to lower case is not 100% safe. Setting names in UI5 are case sensitive. Two names that differ only in case could here be assumed to be the same and a wrong decision could be taken. It's not likely but could happen.

Aren't all callers safe, except for line 854? It rather should try and decapitalize the string after "bind", instead of lower-casing it. If that fails, it might run a second check with the original spelling (it's not guaranteed that the property name started with a lower case letter, so method creation at runtime might not change the case).)

maxreichmann: Thanks for the review. Would you mind to address this in the original PR #480 from which this is based on?

codeworrior: Too late, #480 has been merged already. @d3xter666 can you lease take a look at my comment above?

d3xter666: Thanks @codeworrior!
I'll take care of this

d3xter666: I have created a fix for that: https://github.com/SAP/ui5-linter/pull/506

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,90 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+        messages: [
+          {
+            column: 48,
+            fatal: true,
+            line: 5,
+            message: 'Value of property \'formatter\' is of type \'string\'.',
+            messageDetails: 'Do not use strings for \'formatter\' values in JavaScript and TypeScript. See Best Practices for Developers (https://ui5.sap.com/#/topic/28fcd55b04654977b63dacbee0552712)',
+            ruleId: 'parsing-error',
+            severity: 2,
+          },
```
maxreichmann, line 26: @flovogt @codeworrior This snapshot shows the detection of formatter properties that have a value of type string in bindings (regardless if strings or object literals). This is not a deprecation, but a helpful hint for devs, as the runtime cannot resolve this declaration even with a formatter import.

For the time being, I have thrown a parsing error for such detections because I could not find a more suitable `ruleId`.

What do you think? Should we design a new `ruleId` for this? e.g.: "no-strings-in-formatter-properties" ? However, this would be very specific and not fit into our naming conventions for ruleIds.

d3xter666: Please, take a look at my comment here: https://github.com/SAP/ui5-linter/pull/499#discussion_r1936779441

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,90 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+          {
+            column: 11,
+            fatal: true,
+            line: 9,
+            message: 'Value of property \'formatter\' is of type \'string\'.',
+            messageDetails: 'Do not use strings for \'formatter\' values in JavaScript and TypeScript. See Best Practices for Developers (https://ui5.sap.com/#/topic/28fcd55b04654977b63dacbee0552712)',
+            ruleId: 'parsing-error',
+            severity: 2,
+          },
```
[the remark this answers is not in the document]
maxreichmann: Is the code you're referring to really related? I think the purpose of it is to handle the `events` property inside the binding if it the type is an **object**.

The check responsible for the snapshot handles **string** types for the `formatter` property. This should be detached from the other check. And imo, it's okay to have multiple findings for a single code snippet like this formatter value. Other linters like ESLint can also produce multiple findings.

test/lib/linter/rules/snapshots/NoGlobals.ts.md
codeworrior: > This snapshot shows the detection of formatter properties that have a value of type string in bindings (regardless if strings or object literals).

I struggle with the "regardless". In binding strings, a formatter string IS supported by the runtime, in binding objects it IS NOT.

Therefore, complaining about a use of strings in a binding object does NOT fall into the category "worked in UI5 1.x, but no longer works in UI5 2.x" and therefore is not in the current focus of the linter. It also does not fall into the category of "no-deprecated-api" as it never worked. It'S rather a "unsupported-api-usage" oder "illegal-argument" or whatever.

In binding strings, it's different.

test/fixtures/linter/rules/NoGlobals/FormatterGlobalBindingString.js
```diff
@@ -0,0 +1,18 @@
...
+			value: "{ path: 'invoice>Status', formatter: 'ui5.walkthrough.model.formatter.statusText' }"
+		});
+		input.applySettings({
+			value: "{ path: 'invoice>Status', formatter: 'ui5.walkthrough.model.formatter.statusText' }",
+		});
+
+		// Although the formatter property does not look like a global notation,
+		// it should still be detected if it is a string:
+		const input2 = new Input({
+			value: "{ path: 'invoice>Status', formatter: 'formatter.statusText' }"
```
d3xter666, line 15: In this case, it's clear that there cannot be fromatter like this:
`"{ path: 'invoice>Status', formatter: formatter.statusText }"`, (no quotes around `formatter.statusText`) because the BindingParser/ExpressionParser will throw an exception.
Therefore, the current state is the only possible way to define this.

Why do we need to report this case, given that the `formatter` is imported as a module?

The only reasonable answer would be that the `formatter` will not be supported when defining a string binding.

@codeworrior do I miss something here?

codeworrior: To avoid misunderstandings, we're talking about the following case with quotes around the formatter _value_ and where formatter is the import name of the corresponding module:
```js
sap.ui.define(["some/formatter", ...], function(formatter, ...) {
    new Input({
        value: "{ path: 'invoice>Status', formatter: 'formatter.statusText' }"
    });
});
```
This does NOT work at runtime. The local variable "formatter" (introduced by the sap.ui.define call) is visible only within the scope of the module factory function. The framework code (which parses the binding string) is in a different scope and does not know this variable nor does it have access to it. So, for the framework code, "formatter" looks like any other global name and that's why it should be reported.

codeworrior: I suggested to add this as a fixture mainly for ourselves, so that we don't think this would be a valid usage of a formatter in a binding string.

codeworrior: > The only reasonable answer would be that the formatter will not be supported when defining a string binding.

If you change this slightly to 
> The only reasonable recommendation would be that the formatter will no longer be supported in UI5 2.0 when defining a string binding.

then I agree.

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,86 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+        filePath: 'FormatterGlobalBindingObject.js',
+        messages: [
+          {
+            column: 48,
+            line: 6,
+            message: 'Do not use strings for \'formatter\' values in JavaScript and TypeScript.',
+            messageDetails: 'See Best Practices for Developers (https://ui5.sap.com/#/topic/28fcd55b04654977b63dacbee0552712)',
+            ruleId: 'unsupported-api-usage',
+            severity: 2,
+          },
```
maxreichmann, line 25: @flovogt Please review this rule name.

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,86 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+          {
+            column: 11,
+            line: 15,
+            message: 'Access of global variable \'formatter\' (formatter.statusText)',
+            messageDetails: 'Do not use global variables to access UI5 modules or APIs. See Best Practices for Developers (https://ui5.sap.com/#/topic/28fcd55b04654977b63dacbee0552712)',
+            ruleId: 'no-globals',
+            severity: 2,
+          },
```
maxreichmann, line 81: Lines 58-81:
These findings are the result of detecting strings for formatter values inside a Binding String (previously: "parsing-error", now: "no-globals").

test/fixtures/linter/rules/NoGlobals/FormatterGlobalBindingString.js
d3xter666: Now, I get it! Thanks a lot!

d3xter666 (commented):
LGTM

test/lib/linter/rules/snapshots/NoGlobals.ts.md
KlattG: LGTM

d3xter666 (approved):

test/lib/linter/rules/snapshots/NoGlobals.ts.md
```diff
@@ -4,6 +4,86 @@ The actual snapshot is saved in `NoGlobals.ts.snap`.
...
+    ]
+
+## General: FormatterGlobalBindingString.js
+
+> Snapshot 1
+
+    [
+      {
+        coverageInfo: [],
+        errorCount: 5,
```
matz3, line 38: When looking at the file I would expect 3 findings. To me it looks like there are two duplicate findings.

maxreichmann: As aligned with @codeworrior and approved by @flovogt, we want to detect two cases here:
1. If the `formatter` property is of type **string** inside a **BindingString**
2. If a global notation of the formatter property was used (here: "ui5.walkthrough...")

Since both of these cases apply for this file, there are two findings each.
Nonetheless, we've decided to use the same linter message.

matz3: Both cases should cause a finding, but as they originate from the same place in the code, there is only one finding to be solved. So only one should be reported.

maxreichmann: Maybe we should adjust the linter messages to better emphasize the difference. But both cases should still be reported and not just one imo.

PR #506, opened 2025-01-29
fix: Provide check for alternative names of the properties

d3xter666:
Resolves: https://github.com/SAP/ui5-linter/pull/499#discussion_r1930628950

### DIFF
--- a/src/linter/ui5Types/SourceFileLinter.ts
+++ b/src/linter/ui5Types/SourceFileLinter.ts
@@ -1247,7 +1247,7 @@ export default class SourceFileLinter {
 	 * from there. Directly finding the type of the property is not possible from here as it's
 	 * missing some context.
 	*/
-	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string[] | undefined) {
+	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string[]) {
 		const controlAmbientModule =
 			this.getSymbolModuleDeclaration(this.checker.getTypeAtLocation(node).symbol);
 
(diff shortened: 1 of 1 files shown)


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1241,7 +1247,7 @@ export default class SourceFileLinter {
 	 * from there. Directly finding the type of the property is not possible from here as it's
 	 * missing some context.
 	*/
-	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string | undefined) {
+	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string[] | undefined) {
```
codeworrior, line 1250: I think, with the new line 1182, the `propName` (array) can no longer be `undefined`. But an entry in the array can be the empty string (which then shouldn't match anything)

d3xter666: Ah, thanks!

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1247,7 +1247,7 @@ export default class SourceFileLinter {
 	 * from there. Directly finding the type of the property is not possible from here as it's
 	 * missing some context.
 	*/
-	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string[] | undefined) {
+	#isPropertyBinding(node: ts.NewExpression | ts.CallExpression, propName: string[]) {
```
codeworrior, line 1250: [minor] Optional chaining operator "?" after propName in line 1263 is now also obsolete...

d3xter666: Thanks!

codeworrior (approved):
LGTM

PR #510, opened 2025-01-30
feat: OData implicit globals detection

d3xter666:
JIRA: CPOUI5FOUNDATION-915


src/linter/binding/BindingLinter.ts
```diff
@@ -152,4 +152,29 @@ export default class BindingLinter {
 			this.#context.addLintingMessage(this.#resourcePath, MESSAGE.PARSING_ERROR, {message}, position);
 		}
 	}
+
+	#isExpressionBinding(bindingDefinition: string): boolean {
+		return /^\{:?=/.test(bindingDefinition) && bindingDefinition.endsWith("}");
+	}
+
+	lintPropertyExpression(
```
d3xter666, line 160: It's not possible to use the BindingParser directly as it strips the information of the globals. However, in the runtime code a warning for those gets fired: https://github.com/SAP/ui5-linter/blob/main/src/linter/binding/lib/ExpressionParser.js#L125

The code in the `ExpressionParser` does not significantly differs from the current implementation. It just checks whether those globals are present in the expression

maxreichmann (approved):

PR #513, opened 2025-02-03
refactor: Consider ManagedObject as base class for bind* & applySettings

d3xter666:
fixes: https://github.com/SAP/ui5-linter/pull/480#discussion_r1938892332

### DIFF
--- a/src/linter/ui5Types/SourceFileLinter.ts
+++ b/src/linter/ui5Types/SourceFileLinter.ts
@@ -850,7 +850,7 @@ export default class SourceFileLinter {
 				this.#analyzePropertyBindings(node.arguments[1], ["type", "formatter"]);
 			} else if (symbolName.startsWith("bind") &&
 				nodeType.symbol?.declarations?.some((declaration) =>
-					this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control")) &&
+					this.isUi5ClassDeclaration(declaration, "sap/ui/base/ManagedObject")) &&
 					node.arguments[0] && ts.isObjectLiteralExpression(node.arguments[0])) {
 				// Setting names in UI5 are case sensitive. So, we're not sure of the exact name of the property.
 				// Check decapitalized version of the property name as well.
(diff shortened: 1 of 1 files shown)


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -842,7 +842,7 @@ export default class SourceFileLinter {
 				this.#reportTestStarter(node);
 			} else if (symbolName === "applySettings" &&
 				nodeType.symbol?.declarations?.some((declaration) =>
-					this.isUi5ClassDeclaration(declaration, "sap/ui/core/Control"))) {
+					this.isUi5ClassDeclaration(declaration, "sap/ui/base/ManagedObject"))) {
```
matz3, line 845: Please check for all places. There is still a check against Control a few lines below.

d3xter666: `bind` in here: https://github.com/SAP/ui5-linter/pull/513/files#diff-7bc0071c1214325807d3a82a222b097360e3b0020aae68d0575dbae4aa43b99aR851 reflects the `bind${propertyName}` which, according to the findings in the docs, is being generated at the level where control properties are defined. 
It would be the most appropriate to check against `moduleName`, but still the inheritance factor plays a role, so the exact and correct finding which module is would add complexity and performance penalty.

I agree that `bindProperty` should be checked aggainst `ManagedObject`, because it's defined in there, but the case here is more complex and `sap/ui/core/Control` seems to me a reasonable compromise

matz3: The `bind${propertyName}` method is generated for a ManagedObject sub-class that defines properties in their metadata. Most case are in fact based on sub-classes of Control, but some are also based on Element and some are just sub-classes of ManagedObject (see "View Subclasses" in https://ui5.sap.com/#/api/sap.ui.base.ManagedObject).

As we have now several places for that check, maybe doing it only once would even improve performance.

d3xter666: Thanks!

PR #514, opened 2025-02-03
fix(formatter): Remove duplicate findings

maxreichmann:
+ fix(formatter): Remove duplicate findings
+ docs(Rules.md): Add explanation for `unsupported-api-usage` (new rule introduced by previous formatter change)


src/linter/binding/BindingLinter.ts
```diff
@@ -136,7 +143,9 @@ export default class BindingLinter {
 			}
 		} catch (err) {
 			const message = err instanceof Error ? err.message : String(err);
-			this.#context.addLintingMessage(this.#resourcePath, MESSAGE.PARSING_ERROR, {message}, position);
+			if (err instanceof Error) {
```
matz3, line 146: I think this still comes from the other change about improving the position for errors. I'm not sure whether `err` actually always is an instance of Error, but if this is ensured, we can also adjust the line that handles the message.

test/fixtures/linter/rules/NoGlobals/FormatterGlobalBindingString.js
```diff
@@ -12,7 +12,7 @@ sap.ui.define(
 		// Although the formatter property does not look like a global notation,
 		// it should still be detected if it is a string:
 		const input2 = new Input({
-			value: "{ path: 'invoice>Status', formatter: 'formatter.statusText' }"
+			value: "{ path: 'invoice>Status', formatter: formatter.statusText }"
```
matz3, line 15: This causes a parsing error, which I think isn't the focus within this test fixture.

docs/Rules.md
```diff
@@ -82,3 +83,7 @@ Syntax/parsing errors that appear during the linting process are reported with t
 ## ui5-class-declaration
 
 Checks whether the declaration of UI5 classes is correct. This rule only applies to TypeScript code where built-in ECMAScript classes are used instead of an `.extend()` call.
+
+## unsupported-api-usage
+
+Checks whether used syntax is unsupported by UI5.
```
flovogt, line 89:
```suggestion
Checks whether UI5 API is used correctly, for example in a binding declaration a formatter has to be of type `function` when it's declared in a JavaScript/TypeScript file.
```

flovogt: @klattg please verify

KlattG:
```suggestion
Checks whether the UI5 API is used correctly, for example, whether a formatter in a JavaScript/TypeScript binding declaration is of type `function`.
```

KlattG: @flovogt I'd suggest to write it like this. Is it OK for you?

flovogt: LGTM

matz3 (approved):

d3xter666 (approved):

PR #517, opened 2025-02-06
feat: Detect inline XML fragments & views

d3xter666:
JIRA: CPOUI5FOUNDATION-986


matz3:
This has been addressed via https://github.com/SAP/ui5-linter/pull/519

PR #518, opened 2025-02-07
refactor: Use TypeScript LanguageService

matz3:
Switching from just a `Program` to the `LanguageService` has several benefits:

- Adding/changing source files is possible without having to create a new program manually while keeping all unchanged files cached
- Host implementation is more lightweight
- A shared `LanguageService` can be used to improve performance when linting multiple independent files/projects within the same process/thread. This is currently only used internally by the tests, but a new public API can be easily provided.

### DIFF
--- a/src/linter/linter.ts
+++ b/src/linter/linter.ts
@@ -10,10 +10,11 @@ import {ProjectGraph} from "@ui5/project";
 import type {AbstractReader, Resource} from "@ui5/fs";
 import ConfigManager, {UI5LintConfigType} from "../utils/ConfigManager.js";
 import {Minimatch} from "minimatch";
+import SharedLanguageService from "./ui5Types/SharedLanguageService.js";
 
 export async function lintProject({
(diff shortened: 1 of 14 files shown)


d3xter666 (approved):
LGTM

src/linter/linter.ts
```diff
@@ -10,10 +10,11 @@ import {ProjectGraph} from "@ui5/project";
 import type {AbstractReader, Resource} from "@ui5/fs";
 import ConfigManager, {UI5LintConfigType} from "../utils/ConfigManager.js";
 import {Minimatch} from "minimatch";
+import SharedLanguageService from "./ui5Types/SharedLanguageService.js";
```
codeworrior, line 13: seems that in this file, a type import would be sufficient?

matz3: Correct. Thanks. Earlier I created an instance from `lintFile`, but then moved it out to improve the test performance.

matz3:
Ready for review 👍🏻

d3xter666 (approved):

PR #519, opened 2025-02-07
feat: Detect inline XML fragments & views

d3xter666:
JIRA: CPOUI5FOUNDATION-986


src/linter/ui5Types/TypeLinter.ts
```diff
@@ -135,14 +146,43 @@ export default class TypeChecker {
...
+				linterDone();
+			}
+		}
+
+		// Will eventually produce new JS files for XML views and fragments
+		await lintXml({
+			filePathsWorkspace: this.#filePathsWorkspace,
+			workspace: this.#workspace,
+			context: this.#context,
+		});
```
matz3, line 161: doesn't this lint all files again?

d3xter666: Not all, but just the newly created ones. 
I have refactored it to respect the `xmlCompiledResource` flag: https://github.com/SAP/ui5-linter/pull/519/files#diff-1bc919fd57fdbc38b20c5feaca329ba1764224846e55d4a8b7b654f18e9ca35cR22
In essence, this means that the XML resource has already been linted, so we can skip it.

I was thinking in another direction- to provide a custom AbstractReader/Writer, but we need the "global" ones as inside we stash the transpiled XML->JS resources that are later being used in the SourceFileLinter

matz3:
> **Note**: This change sorts messages in the tests, so that many unrelated test snapshots are also included in here. The reasoning is that XML extraction is not stable, so the order of the messages in different runs/OS might be shuffled.

Currently, the linting results are deterministic and we should stick with that. Only sorting the results when comparing within unit tests will not lead to deterministic results for public consumers.

I suspect that this is caused by a missing sorting of the newly created files. For the initial files, this is done via: https://github.com/SAP/ui5-linter/blob/70ad529dbdfddf0a26b000fdc177e060ff5aacee/src/linter/ui5Types/TypeLinter.ts#L87-L88

src/linter/ui5Types/TypeLinter.ts
```diff
@@ -146,7 +146,43 @@ export default class TypeChecker {
...
+		const virtualXMLResources =
+			await this.#filePathsWorkspace.byGlob("/**/*.inline-*.view{.js,.ts,.js.map}");
```
matz3, line 163: a test case with a `.ts` file would be good to cover the typescript use case

test/lib/linter/rules/snapshots/XMLInJs.ts.md
```diff
@@ -0,0 +1,220 @@
...
+            messageDetails: 'Deprecated test message',
+            ruleId: 'no-deprecated-api',
+            severity: 2,
+          },
+        ],
+        warningCount: 0,
+      },
+    ]
+
+## General: SapUIFragment.js
```
matz3, line 31: Same hint here and for some other files regarding the naming convention.

test/fixtures/linter/rules/XMLInJs/SapUIView.js
```diff
@@ -0,0 +1,38 @@
...
+			xmlns:mvc="sap.ui.core.mvc"
+			xmlns:core="sap.ui.core"> 
+				<Button tap=".sayHello">
+			</mvc:View>`,
+	});
+	oView.placeAt("content");
+	
+	const oView2 = sap.ui.view({
+		type: "XML",
+		viewContent: `<mvc:View
```
matz3, line 17: Please also include a test case where a normal string literal is used within one line. I could reproduce a wrong column information when using this:
`viewContent: '<mvc:View controllerName="ui5app.controller.Home" displayBlock="true" xmlns="sap.m" xmlns:mvc="sap.ui.core.mvc" xmlns:core="sap.ui.core"><Button tap=".sayHello"></mvc:View>'`

matz3 (changes requested):

test/fixtures/linter/rules/XMLInJs/ViewCreate.js
```diff
@@ -0,0 +1,29 @@
...
+		displayBlock="true"
+		xmlns="sap.m"
+		xmlns:mvc="sap.ui.core.mvc"
+		xmlns:core="sap.ui.core"> 
+		  <Button tap=".sayHello">
+		</mvc:View>`,
+		});
+		oView.placeAt("content");
+
+		const oView2 = XMLView.create({
```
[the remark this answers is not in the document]
d3xter666: Thanks!

test/fixtures/linter/rules/XMLInJs/SapUIView.js
d3xter666: Done

test/lib/linter/rules/XMLInJs.ts
```diff
@@ -0,0 +1,5 @@
+import {fileURLToPath} from "node:url";
```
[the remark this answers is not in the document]
d3xter666: Done

test/lib/linter/rules/snapshots/XMLInJs.ts.md
d3xter666: Done

test/fixtures/linter/rules/XMLInJs/SapUIView.js
matz3: To me the column info seems wrong. It is 146 but should be 162, isn't it?

test/fixtures/linter/rules/XMLInJs/FragmentLoad.js
```diff
@@ -0,0 +1,10 @@
+sap.ui.require(["sap/ui/core/Fragment"], async (Fragment) => {
+	const content = await Fragment.load({
+		type: "XML",
+		definition: `<core:FragmentDefinition
```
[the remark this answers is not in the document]
d3xter666: Thanks! Now is fixed

src/linter/ui5Types/TypeLinter.ts
```diff
@@ -146,7 +146,43 @@ export default class TypeChecker {
...
+					// Remove ".map" from path to have it reflect the associated source path
+					resourcePath.slice(0, -4),
+					await resource.getString()
+				);
+			} else {
+				files.set(resourcePath, await resource.getString());
+			}
+		}
+		program = this.#sharedLanguageService.getProgram();
+		checker = program.getTypeChecker();
```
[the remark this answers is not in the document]
d3xter666: Thanks!

test/fixtures/linter/rules/XMLInJs/ViewCreate.js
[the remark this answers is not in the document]
d3xter666: Done

test/fixtures/linter/rules/XMLInJs/SapUIView.js
```diff
@@ -0,0 +1,38 @@
+sap.ui.define([], () => {
+	const oView = sap.ui.view({
+		type: "XML",
+		definition: `<mvc:View
```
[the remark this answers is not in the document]
d3xter666: Not quite sure about that specifically. 
In the documentation I cannot find information that it supports inline definitions at all: https://sapui5.hana.ondemand.com/sdk/#/api/sap.ui%23methods/sap.ui.view

In the openui5 repo I see factories and cannot trace directly the code itself to judge on that.
That's why I leaned on the BLI where we have stated it like this:
- sap.ui.view: Parameter "definition"
- sap.ui.view: Parameter "viewContent"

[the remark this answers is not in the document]
flovogt: `sap.ui.view` supports only "viewContent".

"definition" is only available via the asynchronous `View.create` API

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -103,6 +109,21 @@ export default class SourceFileLinter {
 				this.#reportTestStarter(this.sourceFile);
 			}
 
+			let i = 0;
+			for (const xmlContent of this.#xmlContents) {
+				const fileName = `${this.sourceFile.fileName.replace(/(\.js|\.ts)$/, "")}.inline-${++i}.view.xml`;
```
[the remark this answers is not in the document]

test/fixtures/linter/rules/XMLInJs/ViewCreate.js
```diff
@@ -0,0 +1,29 @@
+sap.ui.define(
+	["sap/ui/core/mvc/View", "sap/ui/core/mvc/XMLView"],
+	async (View, XMLView) => {
+		const oView = await View.create({
+			type: "XML",
```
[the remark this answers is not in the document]
d3xter666: Done

test/fixtures/linter/rules/XMLInJs/SapUIView.js
d3xter666: Thanks! Now it's fixed

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -204,7 +204,8 @@ export default class SourceFileLinter {
 				if (ts.isPropertyAssignment(prop) &&
 					(ts.isIdentifier(prop.name) || ts.isStringLiteralLike(prop.name)) &&
 					ts.isStringLiteralLike(prop.initializer) &&
-					["definition", "fragmentContent", "viewContent"].includes(prop.name.text)) {
+					["definition", "fragmentContent", "viewContent"].includes(prop.name.text) &&
```
matz3, line 207: The same check is done again below in `isXmlInJsCreationCall` isn't it?
I wonder whether it would be easier to combine `isXmlInJsCreationCall` and `getViewTypeFromXMLInJsCreationCall` as they are very similar. Also, as a CallExpression is required, wouldn't it be better to use the CallExpression as indication node to start the checking instead of the isObjectLiteralExpression inside?

matz3: I would expect that this fits into the same category as the partial deprecations, where we perform checks for a specific function call of a module / class.

src/linter/xmlTemplate/linter.ts
```diff
@@ -14,23 +15,43 @@ export default async function lintXml({filePathsWorkspace, workspace, context}:
 	const controllerByIdInfo = new ControllerByIdInfo();
 
 	await Promise.all(xmlResources.map(async (resource: Resource) => {
+		const resourcePath = resource.getPath();
+		const jsPath = resourcePath.replace(/\.xml$/, ".js");
+		const contextMeta = context.getMetadata(jsPath);
+		// This resource has already been transpiled & linted, so no need for duplicate linting
+		if (contextMeta.xmlCompiledResource) {
+			return;
+		}
```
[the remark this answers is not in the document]
d3xter666: In order to pass only the required resources, we need to pass a newly created `AbstractAdapter` with the extracted XMLs.
After that we have couple of options:
- Pass those `AbstractAdapter`s to the `SourceFileLinter`
- Copy newly created files from the fresh `AbstractAdapter`s to the old ones

IMO, this will highly increase the code complexity. Also, for the first case, we might miss eventual findings in multiple files if there are cases in the future like the Async flags

d3xter666:
> > **Note**: This change sorts messages in the tests, so that many unrelated test snapshots are also included in here. The reasoning is that XML extraction is not stable, so the order of the messages in different runs/OS might be shuffled.
> 
> Currently, the linting results are deterministic and we should stick with that. Only sorting the results when comparing within unit tests will not lead to deterministic results for public consumers.
> 
> I suspect that this is caused by a missing sorting of the newly created files. For the initial files, this is done via:
> 
> https://github.com/SAP/ui5-linter/blob/70ad529dbdfddf0a26b000fdc177e060ff5aacee/src/linter/ui5Types/TypeLinter.ts#L87-L88

Actually, the issue is not in the newly created files, because they are not part of the output, but their findings are injected into (maybe) already existing results in certain files. 
To make results deterministic, there should be something general in the output. Now, there's a sorting of the filenames (by filename) and internal messages (by line and the column) before returning and printing the lint messages which is deterministic.

src/linter/LinterContext.ts
```diff
@@ -368,6 +368,15 @@ export default class LinterContext {
 			lintResults.push(this.generateLintResult(resourcePath));
 		}
 
+		// Provide deterministic order of results & messages.
+		// Some of the (virtual) files might be generated during linting
+		// (XML transpilation & extraction) and their order might be non-deterministic.
+		lintResults.sort((a, b) => a.filePath.localeCompare(b.filePath));
+		lintResults.forEach((result) => {
+			result.messages.sort(
```
matz3, line 376: There is already a sorting of messages within `#getFilteredMessages`, that is actually implemented differently. Can't we just always to the sorting there, even if there are no directives? I would prefer having the sorting only in one place.

src/linter/LinterContext.ts
```diff
@@ -368,6 +368,15 @@ export default class LinterContext {
 			lintResults.push(this.generateLintResult(resourcePath));
 		}
 
+		// Provide deterministic order of results & messages.
+		// Some of the (virtual) files might be generated during linting
+		// (XML transpilation & extraction) and their order might be non-deterministic.
+		lintResults.sort((a, b) => a.filePath.localeCompare(b.filePath));
```
matz3, line 374: Sorting by filePath is already done in `src/linter/linter.ts` after the original filePath has been added. I'd prefer to keep it that way, as it might have differences based on the real filesystem paths (e.g. `src` / `test`, `webapp`, ...)

src/linter/xmlTemplate/linter.ts
```diff
@@ -4,12 +4,14 @@ import transpileXml from "./transpiler.js";
...
 import ControllerByIdInfo from "./ControllerByIdInfo.js";
 import {ControllerByIdDtsGenerator} from "./generator/ControllerByIdDtsGenerator.js";
+import {decodedMap, DecodedSourceMap, SourceMapInput, TraceMap} from "@jridgewell/trace-mapping";
 
 // For usage in TypeLinter to write the file as part of the internal WRITE_TRANSFORMED_SOURCES debug mode
 export const CONTROLLER_BY_ID_DTS_PATH = "/types/@ui5/linter/virtual/ControllerById.d.ts";
 
-export default async function lintXml({filePathsWorkspace, workspace, context}: LinterParameters) {
-	const xmlResources = await filePathsWorkspace.byGlob("**/{*.view.xml,*.fragment.xml}");
+export default async function lintXml({filePathsWorkspace, workspace, context, altGlob}: LinterParameters) {
```
matz3, line 12: As this function accepts the common `LinterParameters`, but we only want to implement handling of `altGlob` for the lintXml, I'd suggest to add that parameter only here, not in the general `LinterParameters`. Otherwise one might try to use the parameter with the other functions.

d3xter666: That was exactly my point. It allows extensibility in the future for the other linters.
I'll adjust it as per your comment

d3xter666:
**Note:** Tests would be failing until https://github.com/SAP/ui5-linter/pull/537 gets merged.

The current change allows later injection of messages which causes some of the messages to be shuffled. Sorting of all the messages solves the issue

matz3 (approved):

PR #533, opened 2025-02-11
feat: OData implicit globals detection (JS/TS)

matz3:
Also improves the existing detection in XML which now also detects multiple function calls within one binding.

JIRA: CPOUI5FOUNDATION-1008


src/linter/binding/lib/ExpressionParser.js
```diff
@@ -939,7 +941,8 @@ export default {
 		return {
 			result: {
 				// formatter: formatter,
-				parts: oTokens.parts
+				parts: oTokens.parts,
+				tokens: oTokens.tokens
```
d3xter666, line 945: This is a modification to the original file. Please annotate it

d3xter666 (changes requested):

src/linter/binding/lib/ExpressionParser.js
```diff
@@ -929,7 +929,9 @@ export default {
 
 		/* UI5 LINTER MODIFICATION:
 			Disabled the next few formatter-related lines.
-			The ExpressionParser formatter function is not required for linting purposes
+			The ExpressionParser formatter function is not required for linting purposes.
+
+			Instead, the tokens are returned to allow linting the individual parts of the expression.
```
matz3, line 934: here is the annotation

d3xter666 (dismissed):

d3xter666 (approved):

PR #538, opened 2025-02-14
fix: Prevent false-positive "prefer-test-starter" in testsuite (follow-up)

matz3:
Follow-up of https://github.com/SAP/ui5-linter/pull/498 which only
solved the problem within the test case, but not in an actual project.


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -29,7 +29,7 @@ const QUNIT_FILE_EXTENSION = /\.qunit\.(js|ts)$/;
 
 // This is the same check as in the framework and prevents false-positives
 // https://github.com/SAP/openui5/blob/32c21c33d9dc29a32bf7ee7f41d7bae23dcf086b/src/sap.ui.core/src/sap/ui/test/starter/_utils.js#L287
-const VALID_TESTSUITE = /^\/testsuite(?:\.[a-z][a-z0-9-]*)*\.qunit\.(?:js|ts)$/;
```
codeworrior, line 32: Good catch! 

When the false positive was reported again, I checked the RegExp and found it suspicious. But because the test was against something called "fileName", I mistakenly assumed the "/" to be the culprit :-)

codeworrior (approved):

PR #541, opened 2025-02-14
feat(binding): Add detection for composite bindings (parts)

matz3:
JIRA: CPOUI5FOUNDATION-1012


src/linter/binding/BindingLinter.ts
```diff
@@ -185,6 +190,11 @@ export default class BindingLinter {
 			const bindingInfo = this.#parseBinding(bindingDefinition);
 			if (typeof bindingInfo === "object") {
 				this.#analyzeAggregationBinding(bindingInfo as AggregationBindingInfo, requireDeclarations, position);
+				if (bindingInfo.parts) {
+					for (const part of bindingInfo.parts) {
+						this.#analyzeAggregationBinding(part, requireDeclarations, position);
```
d3xter666, line 195: Can we have a test for that case?
In the coerage results it shows that this line gets visited, but when I debugged it, nothing is actually reported out of it. Maybe, adding a test case would pay in the future

matz3: Yeah it's when there's an expression binding. But in that case we currently only check the tokens. I have to check again, but I think the parts would not contain any information that we want to lint. It mainly contains the `path`, which is currently not checked.

codeworrior: Not sure if I get the scenario right, that you have in mind. But in an expression binding, the nested bindings technically can have the same properties as a normal binding (e.g .types, formatters etc.).  Most often, expressions are based on the raw data, so typically no types or formatters are used. But technically, it's possible.

matz3: Thank you for your feedback. I invested some more time into checking what is possible at runtime and found some more cases that were not covered.

d3xter666 (approved):
LGTM

PR #546, opened 2025-02-17
fix: False-positives for requestCurrencyCodes/requestUnitsOfMeasure

matz3:
Follow-up of https://github.com/SAP/ui5-linter/pull/481

### DIFF
--- a/src/linter/binding/BindingLinter.ts
+++ b/src/linter/binding/BindingLinter.ts
@@ -43,6 +43,7 @@ export default class BindingLinter {
 				this.checkForGlobalReference(eventHandler, requireDeclarations, position);
 			}
 		}
+
 		if (path) {
 			// Check for computed annotations (@@ syntax)
 			const atAtIndex = path.indexOf("@@");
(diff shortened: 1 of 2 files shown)


src/linter/binding/BindingLinter.ts
```diff
@@ -43,8 +43,10 @@ export default class BindingLinter {
 				this.checkForGlobalReference(eventHandler, requireDeclarations, position);
 			}
 		}
-		if (path) {
-			// Check for computed annotations (@@ syntax)
+
+		// Check for computed annotations (@@ syntax)
+		// Exclude built-in paths for currency and units of measure (see ODataMetaModel.getCodeListTerm)
+		if (path && path !== "/##@@requestCurrencyCodes" && path !== "/##@@requestUnitsOfMeasure") {
```
codeworrior, line 49: Syntax-wise, the two built-in functions are not limited to the root path. They are resolved as built-in functions whenever they occur after "@@". See https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/model/odata/v4/ODataMetaModel.js#L1268-L1274.

However, their implementation complains if the path is not the root path - if I get the source right.

codeworrior (dismissed):
LGTM.

src/linter/binding/BindingLinter.ts
matz3: Thanks. I haven't checked the code in that detail, but based on that I would handle the function in general, not only as root path. I've updated the commit.

codeworrior (approved):
LGTM, thanks!

PR #552, opened 2025-02-19
fix: Valid Js syntax after XML transpilation

d3xter666:
Resolves an issue where the tranpiled XML->Js is exported with a syntax error.
Now this is fixed and the export statements are valid.

Currently, this is not an issue for the linter, but in future this might lead to unexpected results if we try to analyze exports.


test/lib/linter/xmlTemplate/snapshots/transpiler.ts.md
```diff
@@ -26,21 +26,21 @@ Generated by [AVA](https://avajs.dev).
         text: "Hello \\"World\\"",␊
     });␊
     ␊
-    export default const oView = new View({␊
+    const oView = new View({␊
```
flovogt, line 29: why not `export new View({`?

d3xter666: Basically, this is the simplest solution that will not break anything.
The impact of this change is just the export of XMLView. 
However, if we want to implement `export default new View...`, then we'll need to also modify the code for variable declaration & initialization which resides in another module and will definitelly imact the other scenarios

matz3 (approved):

PR #559, opened 2025-02-20
fix: False-positives for event handlers in XML

d3xter666:
When preprocessing XML views/fragments via XML templating it's possible to provide bindings fоr every attribute no matter if it's property, aggregation, event, etc.
Therefore, the linter needs to analyze the bindings within events as "normal" bindings and not just like events.

This change ensures that no false-positives are reported in case a valid binding is present.


matz3:
Can you also add a fixture for the `class` attribute? That one is also used in XML templating e.g. `class="{=${device>/system/phone} ? sapMObjectStatusSmall : sapMObjectStatusLarge}"`

src/linter/xmlTemplate/Parser.ts
```diff
@@ -568,8 +569,22 @@ export default class Parser {
...
+					let bindingInfo = null;
+					try {
+						bindingInfo = BindingParser.complexParser(prop.value, null, false, true, true, true);
+						if (typeof bindingInfo === "object") {
+							this.#bindingLinter.lintPropertyBinding(prop.value, this.#requireDeclarations, position);
+						}
+					} catch (err) {
+						const message = err instanceof Error ? err.message : String(err);
+						this.#context.addLintingMessage(this.#resourcePath, MESSAGE.PARSING_ERROR, {message}, position);
+					}
```
matz3, line 583: I don't like that the parser is used here directly again, as it may differ from the parsing within the BindingLinter. Also, the binding is parsed twice, which is not necessary.

What about returning the bindingInfo in `this.#bindingLinter.lintPropertyBinding`?

src/linter/xmlTemplate/Parser.ts
```diff
@@ -568,8 +569,22 @@ export default class Parser {
...
+
 					EventHandlerResolver.parse(prop.value).forEach((eventHandler) => {
-						if (eventHandler.startsWith("cmd:")) {
+						if (eventHandler.startsWith("cmd:") ||
+							(bindingInfo && eventHandler.startsWith("{") && eventHandler.endsWith("}"))) {
```
matz3, line 587: This check can already be done before calling `EventHandlerResolver.parse`, as we should not try to parse bindings as event handlers.

src/linter/xmlTemplate/Parser.ts
d3xter666: Done

src/linter/xmlTemplate/Parser.ts
```diff
@@ -563,11 +563,16 @@ export default class Parser {
...
-					this.#bindingLinter.lintPropertyBinding(prop.value, this.#requireDeclarations, position);
-				} else if (this.#apiExtract.isAggregation(symbolName, prop.name)) {
+				if (this.#apiExtract.isAggregation(symbolName, prop.name)) {
 					this.#bindingLinter.lintAggregationBinding(prop.value, this.#requireDeclarations, position);
 				} else if (this.#apiExtract.isEvent(symbolName, prop.name)) {
+					// In XML templates, it's possible to have bindings in event handlers
+					// We need to parse and lint these as well
+					const bindingInfo =
+						this.#bindingLinter.lintPropertyBinding(prop.value, this.#requireDeclarations, position);
+					if (bindingInfo && prop.value.startsWith("{") && prop.value.endsWith("}")) {
```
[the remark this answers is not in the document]
d3xter666: Please, check the new solution. It verifies if the potential function name is a valid JS identifier.
The pattern is a bit enhanced to support the all the variants of using functions in XML.
Source: https://github.com/SAP/ui5-linter/blob/3abcf563199e11519d996993dab18609da8c1748/src/linter/html/transpiler.ts#L329

matz3:
Please add new cases which previously caused false-positive findings to `test/fixtures/linter/rules/NoGlobals/GlobalEventHandlers.view.xml`.

src/linter/xmlTemplate/Parser.ts
```diff
@@ -563,13 +563,22 @@ export default class Parser {
...
 					EventHandlerResolver.parse(prop.value).forEach((eventHandler) => {
-						if (eventHandler.startsWith("cmd:")) {
+						// Check for a valid function/identifier name
+						// Currently XML views support the following syntaxes that are covered with this pattern:
+						// - myFunction
+						// - .myFunction
+						// - my.namespace.myFunction
+						// - my.namespace.myFunction(arg1, ${i18n>key}, "test")
+						const validFunctionName = /^(\.?[$_\p{ID_Start}](?:[$_\p{ID_Continue}]*))+(\(.*\))?$/u;
+						if (eventHandler.startsWith("cmd:") || !validFunctionName.test(eventHandler)) {
```
matz3, line 581: The `cmd:` check can be done before checking the RegExp.

src/linter/xmlTemplate/Parser.ts
```diff
@@ -563,16 +563,28 @@ export default class Parser {
...
 							// No global usage possible via command execution
 							return;
 						}
+						// Check for a valid function/identifier name
+						// Currently XML views support the following syntaxes that are covered with this pattern:
+						// - myFunction
+						// - .myFunction
+						// - my.namespace.myFunction
+						// - my.namespace.myFunction(arg1, ${i18n>key}, "test")
+						const validFunctionName = /^\.?[$_\p{ID_Start}][$\p{ID_Continue}]*(?:\.[$_\p{ID_Start}][$\p{ID_Continue}]*)*(\(.*\))?$/u;
```
matz3, line 584: As we have a regular expression now to check whether there is a valid function name, why don't we also use it to extract that name instead of checking for it again below? This would reduce the code and ensures consistency.

d3xter666: Done

matz3:
Ready for review

src/linter/xmlTemplate/Parser.ts
```diff
@@ -564,7 +564,9 @@ export default class Parser {
 					line: prop.start.line + 1, // Add one to align with IDEs
 					column: prop.start.column + 1,
 				};
-				if (this.#apiExtract.isAggregation(symbolName, prop.name)) {
+				if (this.#apiExtract.isProperty(symbolName, prop.name)) {
```
d3xter666: This implementation skips "native" properties like "class", for example.
Everything that is not aggregation or event, should be treated as property.

Is there a reason for reverting it back?

matz3: This PR solves false-positives for event handlers. Also checking the other attributes as property binding is a different topic, and mixing both in the same PR makes it hard to judge the differences of findings within projects that this PR creates.

I will create another PR for also checking the other attributes.

d3xter666 (commented):
LGTM, but I don't have rights to approve

src/linter/xmlTemplate/Parser.ts
```diff
@@ -568,18 +569,69 @@ export default class Parser {
...
+						// - myFunction
+						// - .myFunction
+						// - my.namespace.myFunction
+						// - my.namespace.myFunction(arg1, ${i18n>key}, "test")
+						const validFunctionName = /^(\.?[$_\p{ID_Start}][$\p{ID_Continue}]*(?:\.[$_\p{ID_Start}][$\p{ID_Continue}]*)*)\s*(?:\(|$)/u;
+						const match = validFunctionName.exec(eventHandler);
+						if (match) {
+							functionName = match[1];
+						} else if (!eventHandler.includes("(") && !bindingInfo) {
+							// Simple case of a function call without arguments which can
```
codeworrior, line 624: Good catch! Strange usage, but technically possible (unfortunately)

codeworrior (approved):
LGTM

src/linter/xmlTemplate/Parser.ts
matz3: Done via https://github.com/SAP/ui5-linter/pull/575

PR #567, opened 2025-02-25
refactor: Update API usage of sax-wasm

matz3:
- Updating to the new recommended API of sax-wasm
- Reuse parseXML in xml transpiler

### DIFF
--- a/src/utils/xmlParser.ts
+++ b/src/utils/xmlParser.ts
@@ -3,6 +3,7 @@ import {Detail, Reader, SaxEventType, SAXParser, Tag, Text} from "sax-wasm";
 import {finished} from "node:stream/promises";
 import fs from "node:fs/promises";
 import {createRequire} from "node:module";
+import {Readable} from "node:stream";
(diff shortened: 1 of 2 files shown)


src/utils/xmlParser.ts
```diff
@@ -38,27 +39,16 @@ export async function parseXML(
...
-			}
-		}
-	});
+	const webContentStream = Readable.toWeb(contentStream);
+
+	for await (const [event, detail] of saxParser.parse(webContentStream.getReader())) {
+		parseHandler(event, detail);
+	}
+
 	await finished(contentStream);
```
codeworrior, line 53: is finished still required? I would expect that the iterator is exhaustive? The sax-wasm README does not call finished in the ReadableStream example...

BTW: toWeb is still experimental in Node 23 (according to their docs)

matz3: Yeah I guess not but I thought it shouldn't hurt leaving it there. But good point about the experimental API. Then we shouldn't use it, yet.

PR #572, opened 2025-02-27
feat(Autofix): UI5 globals

d3xter666:
This adds the first version of the autofix feature.
A new '--fix' option is added which enables the autofix feature.

It supports fixing usages of UI5 globals with their corresponding UI5
module imports.

JIRA: CPOUI5FOUNDATION-989


.nycrc
```diff
@@ -7,7 +7,8 @@
 		"src/cli/utils/profile.ts",
 		"src/linter/xmlTemplate/lib/JSTokenizer.js",
 		"src/linter/binding/lib/**/*.js",
-		"**/*.d.ts"
+		"**/*.d.ts",
+		"src/autofix/**"
```
d3xter666, line 11: This is needed temporary as currently we haven't implemented all the autofix functionality and it will be done in the next BLIs

src/autofix/autofix.ts
```diff
@@ -0,0 +1,380 @@
...
+			} else if ("namespace" in nodeInfo && nodeInfo.namespace === "sap.ui.controller") {
+				// Check whether sap.ui.controller was used to define or create a controller
+				if (ts.isCallExpression(node.parent) && node.parent.arguments.length === 1) {
+					// Creating a controller instance can't be easily replaced as the new API is async
+					continue;
+				} else {
+					// Controller definition
+					nodeReplacement = `return ${nodeReplacement}.extend`;
+				}
+			}
```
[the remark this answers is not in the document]
d3xter666: Thanks

test/e2e/autofix-e2e.ts
```diff
@@ -0,0 +1,43 @@
+import anyTest, {TestFn} from "ava";
```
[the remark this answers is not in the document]
d3xter666: Now it's refactored to execute a real ui5lint CLI against the troubleshot app

src/autofix/autofix.ts
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	const changeSet: ChangeSet[] = [];
+	let newModuleDeclarations: NewModuleDeclarationInfo[] = [];
+	if (messagesById.has(MESSAGE.DEPRECATED_API_ACCESS)) {
+		newModuleDeclarations = generateSolutionDeprecatedApiAccess(
+			checker, sourceFile, content,
+			messagesById.get(MESSAGE.DEPRECATED_API_ACCESS) as RawLintMessage<MESSAGE.DEPRECATED_API_ACCESS>[],
+			changeSet);
+	}
```
matz3, line 188: In the first iteration, only globals should be replaced, so this logic should not be in here. I'd be fine with removing the code and coming back to the poc branch at https://github.com/SAP/ui5-linter/tree/add-autofix-poc to add the code again once we tackle migration of deprecated API access / AMD migration.

d3xter666: Currently, this code does nothing as we don't have any tests for it.
I can remove it for easier code review, but still have already extracted the code

matz3: It did nothing in our test scenario, but in real projects it would have tried to do things, which is what we don't want to add for now as a first step.

d3xter666: I have cleaned it up

src/autofix/autofix.ts
```diff
@@ -0,0 +1,318 @@
...
+		});
+	}
+
+	const closeParenToken = moduleDeclaration.factory.getChildren()
+		.find((c) => c.kind === ts.SyntaxKind.CloseParenToken);
+	// Factory arguments
+	const syntaxList = moduleDeclaration.factory.getChildren()
+		.find((c) => c.kind === ts.SyntaxKind.SyntaxList);
+	if (!syntaxList) {
+		throw new Error("Invalid factory syntax");
```
matz3, line 255: This error would fail the whole linter run and not apply any fixes.
IMO, the autofix should catch and log errors (maybe only as verbose log for now) but do not fail, as this would also not reveal any messages to the user.

The log message should also contain more information about where the error appeared (at least within which file). Currently there are no verbose logs at all, so it's not possible to see where it appeared from the logs.

src/autofix/autofix.ts
```diff
@@ -0,0 +1,318 @@
...
+		...imports.map((i) => {
+			const identifier = getIdentifierForImport(i);
+			importRequests.get(i)!.identifier = identifier;
+			return identifier;
+		})];
+
+	if (moduleDeclaration.dependencies) {
+		const depsNode = defineCall.arguments[0];
+		const depElementNodes = depsNode && ts.isArrayLiteralExpression(depsNode) ? depsNode.elements : [];
+		const lastElement = depElementNodes[depElementNodes.length - 1];
```
matz3, line 234: this fails if `depsNode` is empty, as there is no element at all. Please add more test fixtures to cover the non-straight-forward cases. Currently we only have two fixtures (+ the e2e test)

src/autofix/autofix.ts
```diff
@@ -0,0 +1,318 @@
...
+	return applyChanges(content, changeSet);
+}
+
+function addDependencies(
+	defineCall: ts.CallExpression, moduleDeclarationInfo: ExistingModuleDeclarationInfo,
+	changeSet: ChangeSet[], resourceMetadata: LintMetadata
+) {
+	const {moduleDeclaration, importRequests} = moduleDeclarationInfo;
+
+	const existingImportModules = resourceMetadata.transformedImports.get("sap.ui.define") ?? [];
```
matz3, line 206: There is no guarantee that `transformedImports` is filled. If the transforming failed for whatever reason, we do not have this information available.

src/autofix/solutions/noGlobals.ts
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+		if (defineCall === undefined) {
+			defineCall = null;
+		}
+		let moduleDeclaration;
+		if (defineCall) {
+			if (!moduleDeclarations.has(defineCall)) {
+				moduleDeclarations.set(defineCall, {
+					moduleDeclaration: parseModuleDeclaration(defineCall.arguments, checker),
```
matz3, line 95: `parseModuleDeclaration` throws errors in case it does not understand the declaration. This case should be handled in here.

src/autofix/autofix.ts
d3xter666: In this case, failing is the best thing that could happen. It's completelly invalid JS syntax.
A `SyntaxList` is the args of a function. There could be an arrow function with no arguments. And still, if there are braces, it would have an empty `SyntaxList`. This would fail only if we have the following:
```js
 => {}
```
which is eitherways syntax error in JS.

I'll try to enhance the tests and provide more examples to cover all the cases.

This check is only needed for TS validation purposes

src/autofix/autofix.ts
d3xter666: JS files would always go through the transformer, therefore this would be set. TypeScript would have catched this one.
Only TS files would fail, but we do not support them for now either ways.

I'll add a guard about it

matz3: The transformation for JS files *can* fail in certain cases (currently for example when de-structuring is used within factory arguments) and the error would be caught, so the file remains as it is.

d3xter666: Can you provide an example?
I'm currently trying with the following and it's parsed successfully:

```js
sap.ui.define(["sap/m/Button", "sap/m/library"], function(ButtonRenamed, {BackgroundDesign}) {
```

matz3: I sent you the file via chat

d3xter666: Thanks!

src/autofix/autofix.ts
d3xter666: This is not possible. Please take a look at line #231
This is the case where we are ensured that there are dependencies already and we take the last one

test/lib/autofix/snapshots/autofix.ts.md
```diff
@@ -10,7 +10,20 @@ Generated by [AVA](https://avajs.dev).
 
     [
       {
-        coverageInfo: [],
+        coverageInfo: [
+          {
+            category: 1,
+            column: 20,
+            line: 2,
+            message: 'Unable to analyze this method call because the type of identifier in "jQuery("#container-todo---app--avatar-profile")"" could not be determined',
```
matz3, line 18: there seems to be something wrong here. jQuery is the 3rd parameter in the list, which corresponds to "sap/m/Avatar".

src/autofix/solutions/noGlobals.ts
d3xter666: Done

src/autofix/autofix.ts
d3xter666: Ah, thanks!

test/lib/autofix/snapshots/autofix.ts.md
d3xter666: Now, it's fixed. Thanks!

d3xter666: For this specific case, however, the test is intended to throw an exception at `parseModuleDeclaration` (src/autofix/solutions/noGlobals.ts). So, the module is not parsed and fixed.
Please, take a look at the exception message here: https://github.com/SAP/ui5-linter/pull/572/files#diff-0915772a0290304ddfecbbbfff5a184d8dcefb5de22f1fb3e7575d3d5bca90a7R29 
and also to the test case here: https://github.com/SAP/ui5-linter/pull/572/files#diff-121046e7da30b2d272aadc30054758852d6f4c07b304aea853f8d3ef0dcd89e2

src/autofix/solutions/noGlobals.ts
```diff
@@ -0,0 +1,131 @@
...
+		if (defineCall) {
+			if (!moduleDeclarations.has(defineCall)) {
+				try {
+					moduleDeclarations.set(defineCall, {
+						moduleDeclaration: parseModuleDeclaration(defineCall.arguments, checker),
+						importRequests: new Map(),
+					});
+				} catch (err) {
+					const message = err instanceof Error ? err.message : String(err);
+					context.addLintingMessage(sourceFile.fileName, MESSAGE.PARSING_ERROR, {message}, position);
```
matz3, line 103: Reporting such issues only here during autofix, but not within the AMD transpiler (where the same function is used) does not sound consistent to me.

Those errors should rather also just be logged (verbose) and ignored, like in the AMD transpiler. The exceptions do not always mean that something in the code is wrong. It's rather that we do not fully understand it, so we can't transform the module or derive information from it.

My idea would be that we skip the autofix for the files where something like this happens, and log it accordingly (verbose).

We should think about introducing an internal flag (env var) to enable throwing or reporting errors that _might_ indicate that something with our code is wrong or could be improved. But I don't think this information is helpful to end-users, so we should generally hide it.

src/autofix/autofix.ts
```diff
@@ -0,0 +1,334 @@
...
+	const compilerHost = createCompilerHost(sourceFiles);
+	const program = createProgram(resourcePaths, compilerHost);
+
+	const checker = program.getTypeChecker();
+	const res: AutofixResult = new Map();
+	for (const [resourcePath, sourceFile] of sourceFiles) {
+		const newContent = applyFixes(checker, sourceFile, resourcePath, resources.get(resourcePath)!, context);
+		if (newContent) {
+			res.set(resourcePath, newContent);
+		}
```
matz3, line 159: We should catch any error when trying to apply an autofix and skip the file in that case. I'm not sure about how this should be logged then (verbose/warning/error?), but we should definitely not halt the execution, as this prevents the user from seeing any finding or getting files fixed at all.

test/lib/autofix/utils.ts
```diff
@@ -1,38 +1,38 @@
...
-test("getIdentifierForImport: Relative library module (one folder up)", (t) => {
-	t.is(getIdentifierForImport("../library"), "library");
+test("resolveUniqueName: Relative library module (one folder up)", (t) => {
+	t.is(resolveUniqueName("../library"), "library");
 });
 
-test("getIdentifierForImport: Controller module", (t) => {
-	t.is(getIdentifierForImport("./controller/App.controller"), "AppController");
+test("resolveUniqueName: Controller module", (t) => {
+	t.is(resolveUniqueName("./controller/App.controller"), "App_controller");
```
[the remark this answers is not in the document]
d3xter666: Done with https://github.com/SAP/ui5-linter/pull/572/commits/4060fbd7715520994af12e60b83e141fd39fa844

src/autofix/autofix.ts
```diff
@@ -0,0 +1,304 @@
...
+	for (const msg of resource.messages) {
+		if (msg.fixHints?.moduleName && conditionalModuleAccess.has(msg.fixHints.moduleName)) {
+			// Skip messages with conditional fixHints
+			continue;
+		}
+		if (!messagesById.has(msg.id)) {
+			messagesById.set(msg.id, []);
+		}
+		messagesById.get(msg.id)!.push(msg);
+	}
```
d3xter666, line 264: Isn't this the same as `getAutofixMessages()`?

d3xter666 (commented):
LGTM

src/autofix/autofix.ts
matz3: Oh yeah, I think I missed to also refactor this part. Thanks.

matz3: Done

src/linter/ui5Types/amdTranspiler/pruneNode.ts
```diff
@@ -21,11 +21,13 @@ export class UnsafeNodeRemoval extends Error {
  */
 export default function (node: ts.Node) {
 	let nodeToRemove: ts.Node | undefined = node;
+	let greatestRemovableNode: ts.Node | undefined = undefined;
```
matz3, line 24: @d3xter666 is this change needed?

d3xter666: To bo honest, I really don't understand why this has been introduced. I checked the code and it seems redundant now. Maybe, it's just a leftover

matz3: Done

d3xter666 (commented):
LGTM

matz3 (approved):
LGTM

PR #574, opened 2025-02-28
fix: XML namespace handling

matz3:
The stack of XML namespaces was not being properly maintained which
resulted into wrong assumptions about controls in some scenarios when
multiple default namespaces were present in the XML.


src/linter/xmlTemplate/Parser.ts
```diff
@@ -297,13 +297,13 @@ export default class Parser {
 				this._addNamespace({
 					localName: null,
 					namespace: attrValue,
-				}, this.#nodeStack.length);
+				}, this.#nodeStack.length + 1); // +1 as the currently created node is not yet on the stack
```
codeworrior, line 300: if the idea is that `level` counts from 1, I'm fine with this change.- Alternatively, if `level` should correspond to the node's position in the `#nodeStack`, `popTag` could just remove `level - 1`(or swap L171 and L172 and adapt the constant in L178).

codeworrior (approved):
LGTM

src/linter/xmlTemplate/Parser.ts
matz3: Thanks. That's a way better solution.

PR #593, opened 2025-03-06
refactor: Align identifiers name creation

d3xter666:
This change introduces a new module for creation of unique names.
It's purpose it to align varable naming inside the project.


matz3 (approved):

src/linter/ui5Types/utils/UniqueNameCreator.ts
```diff
@@ -0,0 +1,219 @@
...
+	"if",
+	"implements",
+	"import",
+	"in",
+	"instanceof",
+	"int",
+	"interface",
+	"let",
+	"long",
+	"native",
```
codeworrior, line 44: If we want to fix the warning,. we could remove certain identifiers. According to 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#future_reserved_words_in_older_standards, the identifier in question was only reserved from EcmaScript 1 to EcmaScript 3.

d3xter666: fixed with https://github.com/SAP/ui5-linter/pull/595

PR #595, opened 2025-03-06
refactor: Cleanup redundant reserved words

d3xter666:
Cleanup redundant reserve words for the variable name creator. These words have been reserved in older versions of 
 ECMAScript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#future_reserved_words_in_older_standards


matz3 (commented):
LGTM

codeworrior (approved):

Issue #602, opened 2025-03-12
Bootstrap attribute "logLevel" should be reported as a warning finding


flovogt:
Hi team,

for following HTML file we should report a warning finding to use `data-sap-ui-log-level` instead of `data-sap-ui-logLevel`. Documented here https://sdk.openui5.org/topic/91f2d03b6f4d1014b6dd926db0e91070.

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>MyApp</title>
    <script
      id="sap-ui-bootstrap"
      src="../resources/sap-ui-core.js"
      data-sap-ui-theme="sap_horizon"
      data-sap-ui-resource-roots='{
         "myapp": "./",
      }'
      data-sap-ui-on-init="module:luigi/demo/sample2/index"
      data-sap-ui-language="en"
      data-sap-ui-logLevel="none"
      data-sap-ui-async="true"
      data-sap-ui-compat-version="edge"
    ></script>
  </head>
  <body class="sapUiBody">
    <div
      data-sap-ui-component
      data-name="myapp"
      data-id="container"
      data-settings='{"id" : "app"}'
      id="content"
    ></div>
  </body>
</html>
```

Currently no finding is reported at all.

PR #603, opened 2025-03-13
fix: Detect outdated notation for bootstrap attribute log-level

flovogt:
Fixes https://github.com/SAP/ui5-linter/issues/602


flovogt:
@SAP/ui5-foundation Any recommendation if we should add the bootstrap attribute only to one html page and not to every test page?

matz3 (dismissed):
LGTM. I don't mind about adding it to every page.

PR #624, opened 2025-04-11
docs(README.md): Add documentation about usage of explanation comments for directives

flovogt:


README.md
```diff
@@ -348,10 +348,22 @@ An explanation why a rule is disabled can be added after the rule name; it must
...
 * `<!-- ui5lint-disable-next-line -->`
 
-Note that in XML and HTML files, depending on your browser runtime, the use of double-hyphen `--` is usually not allowed, making it impossible to add explanations as part of the directive. Please use separate comment blocks for this purpose.
+Note that in XML and HTML files, depending on your browser runtime, the use of double-hyphen `--` is usually not allowed, making it impossible to add explanations as part of the directive. Please use a separate comment statement in the same line of code for this purpose.
+
+* `<!-- ui5lint-disable-next-line no-deprecated-api --> <!-- explanation -->`
 
 Also, since comments can't be placed inside XML and HTML tags, you might need to use `ui5lint-disable` and `ui5lint-enable` directives outside of the tags instead of `ui5lint-disable-line` or `ui5lint-disable-next-line`.
 
+* `<Text tap=".onTap"/> <!-- ui5lint-disable-line no-deprecated-api -->`
```
RandomByte, line 357: I think this example doesn't go well with the statement right before it, saying you should use `ui5lint-disable` instead of `ui5lint-disable-line`. Maybe move this example up to the first list of XML/HTML examples?

RandomByte (changes requested):

README.md
flovogt: `you might need to use` is not `you should use` :-D But yes, you are right. I'll move it to the top.

flovogt: Done

RandomByte (dismissed):

README.md
```diff
@@ -347,11 +347,22 @@ An explanation why a rule is disabled can be added after the rule name; it must
 
 * `<!-- ui5lint-disable no-deprecated-api -->`
 * `<!-- ui5lint-disable-next-line -->`
+* `<Text tap=".onTap"/> <!-- ui5lint-disable-line no-deprecated-api -->`
 
-Note that in XML and HTML files, depending on your browser runtime, the use of double-hyphen `--` is usually not allowed, making it impossible to add explanations as part of the directive. Please use separate comment blocks for this purpose.
+Note that in XML and HTML files, depending on your browser runtime, the use of double-hyphen `--` is usually not allowed, making it impossible to add explanations as part of the directive. Please use a separate comment statement in the same line of code for this purpose.
```
KlattG, line 352:
```suggestion
In XML and HTML files, depending on your browser runtime, the use of a double-hyphen `--` is usually not allowed, making it impossible to append an explanation to your directive in the same way as shown above for JavaScript / TypeScript files. You need to insert a separate comment statement in the same line of code for this purpose:
```

KlattG (changes requested):
I'd slightly rephrase this to highlight the difference compared to the JS/TS case.

RandomByte (approved):

PR #638, opened 2025-04-23
ci: Use least privilege principle for permissions

matz3:
Declares only the required permissions per workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
     branches:
       - main
 
+# No permissions are required for this workflow
+permissions: {}
+
 jobs:
   test:
     name: General checks, tests and coverage reporting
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,9 @@ on:
     branches:
       - main
 
+# No permissions are required for this workflow
+permissions: {}
```
flovogt, line 11: is the usage of `{}` required? Is an empty value not sufficient, like usually used in `workflow_dispatch:`

matz3: > You can use the following syntax to disable permissions for all of the available permissions:
> ```yaml
> permissions: {}
> ```

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

flovogt: thx for clarification

flovogt (approved):

PR #614, opened 2025-03-28
feat: Autofix deprecated jQuery.sap APIs

d3xter666:
Migrate deprecated legacy jQuery.sap APIs to new recommended APIs.

Fixes: https://github.com/SAP/ui5-linter/issues/520
Fixes: https://github.com/SAP/ui5-linter/issues/521
Fixes: https://github.com/SAP/ui5-linter/issues/522
Fixes: https://github.com/SAP/ui5-linter/issues/524
Fixes: https://github.com/SAP/ui5-linter/issues/525
Fixes: https://github.com/SAP/ui5-linter/issues/527
Fixes: https://github.com/SAP/ui5-linter/issues/528
Fixes: https://github.com/SAP/ui5-linter/issues/529
Fixes: https://github.com/SAP/ui5-linter/issues/531
Fixes: https://github.com/SAP/ui5-linter/issues/532
Fixes: https://github.com/SAP/ui5-linter/issues/542
Fixes: https://github.com/SAP/ui5-linter/issues/543
Fixes: https://github.com/SAP/ui5-linter/issues/555
Fixes: https://github.com/SAP/ui5-linter/issues/561
Fixes: https://github.com/SAP/ui5-linter/issues/562
Fixes: https://github.com/SAP/ui5-linter/issues/563
Fixes: https://github.com/SAP/ui5-linter/issues/589
Fixes: https://github.com/SAP/ui5-linter/issues/590

JIRA: CPOUI5FOUNDATION-990

PR #659, opened 2025-05-14
refactor: Extract code in util to detect expected value from an expression

d3xter666:
`Log.debug/warn/info/etc` returns void, but the legacy `jQuery.log.debug` was returning `this`. The core functionality is the same, except for the return value. This change, fixes some of the cases where return value is expected. Those cases we cannot migrate

fixes: https://github.com/SAP/ui5-linter/pull/614#discussion_r2087461807


matz3:
https://github.com/SAP/ui5-linter/blob/1dda01216deb7dec8c3dd1dd83712ce5166172f3/test/lib/autofix/snapshots/autofix.fixtures.ts.md#L4691-L4692 also shows a case which could replaced but currently isn't.

I'd prefer that we don't use `fix` / `feat` in this case, as there is no release to which this would apply. So for end-users, this is not relevant, IMO.

Also, isn't the commit message title describing the opposite? In this case here we now migrate more calls than before, but the calls that are already migrated are correct, right?

matz3:
As mentioned in https://github.com/SAP/ui5-linter/pull/614#discussion_r2087461807 I would suggest to extract the logic into a util and test it in an isolated scenario (without autofix, like the other ui5Types/utils are tested).

matz3 (approved):

PR #660, opened 2025-05-14
feat: Autofix deprecated class properties in XML

RandomByte:
JIRA: CPOUI5FOUNDATION-994


src/autofix/solutions/generateXmlSolutionDeprecatedApi.ts
```diff
@@ -0,0 +1,67 @@
...
+				action: ChangeAction.REPLACE,
+				start: toPosition(attr.name.start),
+				end: toPosition(attr.name.end),
+				value: classPropertyToBeUsed,
+			});
+		} else {
+			changeSet.push({
+				action: ChangeAction.DELETE,
+				start: toPosition(attr.name.start),
+				end: toPosition(attr.value.end) + 1,
```
matz3, line 57: the `+1` might be wrong if the attribute value is not within quotes

PR #661, opened 2025-05-15
refactor(SourceFileLinter): Extract common logic to utils

matz3:
Extracts common logic into utils and introduces a new
getUi5TypeInfoFromSymbol function which might be useful in future for
usage in the autofix context.


matz3:
Let's split this up into multiple PRs. For now, this PR only extracts some logic without introducing larger changes.

Issue #668, opened 2025-05-27
ui5lint -fix replacing a global in a delete expression results in syntactically invalid code


codeworrior:
## Expected Behavior

When the linter is executed with `--fix` option, the applied replacements create code that is syntactically correct.

## Current Behavior

When a global is replaced in the expression part of a delete expression, the result is not valid strict mode code.

**Before**:
```js
delete sap.ushell.Container;
```
**After**:
```js
delete Container;  // SyntaxError: Delete of an unqualified identifier in strict mode
```

## Steps to Reproduce the Issue

1. Run ui5lint --fix in the sap.ushell library
2.  check the /test-resources/sap/ushell/test/appRuntime/ui5/services/Container.qunit.js module, line 55

## Context

- UI5 linter version: **`1.13.1 (with 1.136.0 deprecation baseline)`**
- Node.js version: **`20.18.0`**
- npm version: `10.8.2`
- OS/Platform: `unknown`

## Log Output / Stack Trace

Runtime Browser Log:
```
Uncaught SyntaxError: Delete of an unqualified identifier in strict mode. (at Container.qunit.js:56:20)
```

PR #669, opened 2025-05-27
fix(autofix): Prevent invalid delete expressions

matz3:
Check for usage of "delete" keyword with the module export, which is not
fixable as only an identifier would remain after the autofix, which
would not be valid.

Fixes: https://github.com/SAP/ui5-linter/issues/668


d3xter666 (approved):
LGTM

PR #672, opened 2025-05-27
refactor: Introduce Ui5TypeInfo

matz3:
Introduces a Ui5TypeInfo which might be especially useful in future for
usage in the autofix context.

The info will be collected when checking for deprecations and attached
to the raw messages. For now, the information is only available
internally but exposing it via the linter messages might be done later
to allow consumers to use it.

JIRA: CPOUI5FOUNDATION-1073

### DIFF
--- a/src/linter/ui5Types/utils/utils.ts
+++ b/src/linter/ui5Types/utils/utils.ts
@@ -346,3 +346,67 @@ export function extractSapUiNamespace(symbolName: string, moduleDeclaration: ts.
 		return namespace.join(".");
 	}
 }
+
+export enum Ui5TypeInfoKind {
+	Module,
+	Global,
+}
+
+export type Ui5TypeInfo = Ui5ModuleTypeInfo | Ui5GlobalTypeInfo;
+
+interface Ui5ModuleTypeInfo {
+	module: string; // module name (slash separated)
+	export: string; // complete export name with namespaces
+	basename?: string; // local name without namespaces
+	name?: string; // e.g. DataType name
+	kind: Ui5TypeInfoKind.Module;
+}
+
+interface Ui5GlobalTypeInfo {
+	namespace: string;
+	kind: Ui5TypeInfoKind.Global;
+}
+
+/**
+ * Extracts module / global type information from UI5 symbols.
+ */
+export function getUi5TypeInfoFromSymbol(symbol: ts.Symbol): Ui5TypeInfo | undefined {
+	let currentNode: ts.Node | undefined = symbol.valueDeclaration;
+	if (!currentNode) {
+		return undefined;
+	}
+	const name = symbol.name;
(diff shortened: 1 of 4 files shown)


RandomByte (approved):

src/linter/ui5Types/utils/utils.ts
```diff
@@ -346,3 +346,67 @@ export function extractSapUiNamespace(symbolName: string, moduleDeclaration: ts.
...
+export enum Ui5TypeInfoKind {
+	Module,
+	Global,
+}
+
+export type Ui5TypeInfo = Ui5ModuleTypeInfo | Ui5GlobalTypeInfo;
+
+interface Ui5ModuleTypeInfo {
+	module: string; // module name (slash separated)
+	export: string; // complete export name with namespaces
```
codeworrior, line 359: **TL;DR:** maybe you should add some marker or syntax to distinguish between named exports and properties of the default export.

**Details**
The api.json also uses the combination of module/export incl. namespaces.  But we meanwhile learned that there's some fuzziness in this storage. It does not allow to distinguish between named exports and properties of the default export. For JavaScript replacements, this doesn't matter. But when we want to autofix TypeScript in future, it makes a difference.

Examples
* Button:
  * module: sap/m/Button 
  * export: (default)
  * access path after export: (none)
* Button.extend:
   * module: sap/m/Button
   * export: (default)
   * access path after export: "extend"
* MessageBox.Action
   * path sap/m/MessageBox
   * export: (default)
   * access path after export: "Action"
 * ButtonType
   * module: sap/m/library
   * export: ButtonType
   * access path after export (none)
 * ButtonType.Default
   * module: sap/m/library
   * export: ButtonType
   * access path after export "Default"

As of now, the information in the api.json can be interpreted correctly "by applying conventions":
* anything with a non-empty `export` name in a */library module is a named export
* an entity with a non-empty `export` that is "type only" (interface, typedef) is a named export.
* everything else with a non-empty `export` is a property of the default export 

But while working on the SeamlessWebComponent integration, we came across another kind of module that uses named exports in TypeScript and which needs special handling. There's a module called "package module". It has a similar role like the library.js, but doesn't form a library (multiple of these package modules can exist in a single project). The module contains for example interfaces, types and enums introduced by a web component package and exposes them in the TS world as named exports. 

To keep changes to the api.json small, we'll introduce a marker "namedExport:boolean". The default is calculated according to the heuristics above and we'll set it explicitly to true for the entities in the package modules.

matz3: Thanks, that's good input 👍🏻
The current state is not ideal, and I also wanted to add tests to verify the result is as expected for certain edge cases.
I want to clarify with @RandomByte whether we should still merge this state now or do the improvements first.

PR #671, opened 2025-05-27
feat: Autofix Core APIs (base)

d3xter666:
JIRA: CPOUI5FOUNDATION-991

Fixes: https://github.com/SAP/ui5-linter/issues/619

Issue #619, opened 2025-04-03
Missing autofix for methods related to deprecated Core APIs


flovogt:
# Missing autofix for methods related to deprecated Core APIs

Methods available on `sap.ui.getCore().*` are deprecated. Therefore an autofix should be offered by UI5 linter.

For a detailed overview about deprecated APIs, see [Deprecated Core API](https://sdk.openui5.org/#/topic/798dd9abcae24c8194922615191ab3f5)

flovogt:
Will be implemented in CPOUI5FOUNDATION-991.

RandomByte:
| Legacy API | Replacement | Additional Information |
|--------|--------|--------|
| [`applyChanges()`](https://ui5.sap.com/#/api/sap.ui.core.Core%23methods/applyChanges) | _Removed without replacement_ | |
| [`applyTheme(sThemeName, sThemeBaseUrl?)`](https://ui5.sap.com/#/api/sap.ui.core.Core%23methods/applyTheme) | `sap/ui/core/Theming.setTheme(sTheme)` | Migrate only if second parameter is omitted? Also: The new API will no longer support providing a new theme root via an "@" symbol. In URL parameters, usage of the "@" symbol is still available. |
| [`attachControlEvent(fnFunction, oListener?)`](https://ui5.sap.com/api/sap.ui.core.Core#methods/attachControlEvent) | _Removed without replacement_  |  | 
| [`attachFormatError(oData?, fnFunction, oListener?)`](https://ui5.sap.com/#/api/sap.ui.core.Core%23methods/attachFormatError) | `ManagedObject#attachFormatError(oData?, fnFunction, oListener?)`| Unclear which managed object to use. Maybe no migration possible? |
| [`attachInit(fnFunction)`](https://ui5.sap.com/#/api/sap.ui.core.Core%23methods/attachInit) | `Core#ready(fnReady?)` |
| | |

PR #675, opened 2025-05-28
feat: Autofix Core API (complex cases)

d3xter666:
JIRA: CPOUI5FOUNDATION-991

Enhances:
https://github.com/SAP/ui5-linter/pull/671
https://github.com/SAP/ui5-linter/pull/675

Enables `getLibraryResourceBundle` migration by checking whether `sLibrary` (the first argument) is an actual SAPUI5 library. The check is performed as follows:
- check for library existence in TS definitions for SAPUI5
- Check project's `libs` in `manifest.json`
- If the current project is a library

Reolves partially: https://github.com/SAP/ui5-linter/issues/619


src/linter/ui5Types/fixHints/CoreFixHintsGenerator.ts
```diff
@@ -313,10 +313,40 @@ export default class CoreFixHintsGenerator {
...
+
+	isPotentialUi5Library(node: ts.Node): boolean {
+		if (!("text" in node) || typeof node.text !== "string") {
+			return false;
+		}
+
+		const potentialLibNamespace = node.text;
+
+		// Check libs from the manifest
+		const manifest = JSON.parse(this.manifestContent ?? "{}") as SAPJSONSchemaForWebApplicationManifestFile;
```
matz3, line 332: Can't we just re-use the existing logic that extracts manifest.json information instead of doing it again here?
(see https://github.com/SAP/ui5-linter/blob/7b1e6dd11d06c22a200b381ee08a4127431b22de/src/linter/lintWorkspace.ts#L173)

We could change this to a broader scope, like `getInformationFromManifest`, and then returning an object with the relevant details.

src/linter/ui5Types/fixHints/CoreFixHintsGenerator.ts
```diff
@@ -313,10 +313,40 @@ export default class CoreFixHintsGenerator {
...
+		const manifestLibs = manifest["sap.ui5"]?.dependencies?.libs ?? {};
+
+		if (manifestLibs[potentialLibNamespace]) {
+			return true;
+		} else if (manifest?.["sap.app"]?.id === potentialLibNamespace) {
```
matz3, line 337: This check should only be done when the project is a library.

src/linter/ui5Types/fixHints/CoreFixHintsGenerator.ts
d3xter666: I just split the existing solution into separate PRs and now I'll do the refactorings

matz3: Ah, sorry for the premature review 😄

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +114,334 @@ t.declareModule("sap/ui/core/Core", [
...
+		// Note that alternative replacement Component.get is meanwhile deprecated, too
 		t.method("getComponent", accessExpressionFix({
 			moduleName: "sap/ui/core/Component",
 			propertyAccess: "getComponentById",
 		})),
+		// Individual arguments must be mapped to "options" object
+		// The new API has no sync loading option, replacement is only safe when the options contain async:true
+		t.method("loadLibrary", callExpressionGeneratorFix({
+			moduleName: "sap/ui/core/Lib",
+			propertyAccess: "load",
```
RandomByte, line 126: This parameter is not used by the generator-fix. This is an issue in the typings that I just corrected in https://github.com/SAP/ui5-linter/pull/714

```suggestion
```

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +114,334 @@ t.declareModule("sap/ui/core/Core", [
...
+								ctx.json![name] = node.initializer.text;
+							}
+						}
+					});
+					if (ts.isStringLiteralLike(arg1)) {
+						ctx.json.name = arg1.text;
+					}
+					return asyncOption;
+				} else if (arg2?.kind === SyntaxKind.TrueKeyword) {
+					if (ts.isStringLiteralLike(arg1)) {
```
RandomByte, line 151: What if it's not a string but an identifier? Shouldn't that be fine as well?

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +114,334 @@ t.declareModule("sap/ui/core/Core", [
...
+						ctx.json.name = arg1.text;
```
RandomByte, line 152: You might want to consider using `getFullText()` which also includes trivia like comments. See for example the StringLiteral here: https://ts-ast-viewer.com/#code/GYexAoHoCoAIBcCmBnetqVgIiarsBKIA

Definitely needs some test cases though, so I'm fine with skipping this for now if its too much work.

```suggestion
						ctx.json.name = arg1.getFullText();
```

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +114,334 @@ t.declareModule("sap/ui/core/Core", [
...
+					ts.forEachChild(arg2, function (node: ts.Node) {
+						if (ts.isPropertyAssignment(node) && ts.isIdentifier(node.name) && node.name.text === "async") {
+							if (node.initializer.kind === SyntaxKind.TrueKeyword) {
+								asyncOption = true; // Migration is possible, async loading is enabled
+							}
+						}
+
+						if (ts.isPropertyAssignment(node) && ts.isIdentifier(node.name)) {
+							const name = node.name.text;
+							if (allowlistProps.includes(name) && ts.isStringLiteralLike(node.initializer)) {
```
RandomByte, line 141: Similar to below, I guess we could also handle identifiers here?

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +114,334 @@ t.declareModule("sap/ui/core/Core", [
...
+								ctx.json![name] = node.initializer.text;
```
RandomByte, line 142: See my comment below, maybe we can preserve trivia here as well.

```suggestion
								ctx.json![name] = node.initializer.getFullText();
```

RandomByte (changes requested):

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
d3xter666: Thanks a lot for your input!

src/linter/ui5Types/TypeLinter.ts
```diff
@@ -120,12 +120,15 @@ export default class TypeLinter {
 				continue;
 			}
 			let manifestContent;
+			let res;
```
RandomByte, line 123: I think we can revert this change the how it was before

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +113,370 @@ t.declareModule("sap/ui/core/Core", [
...
+		// Note that alternative replacement Component.get is meanwhile deprecated, too
 		t.method("getComponent", accessExpressionFix({
 			moduleName: "sap/ui/core/Component",
 			propertyAccess: "getComponentById",
 		})),
+		// Individual arguments must be mapped to "options" object
+		// The new API has no sync loading option, replacement is only safe when the options contain async:true
+		t.method("loadLibrary", callExpressionGeneratorFix<{json: string}>({
+			moduleName: "sap/ui/core/Lib",
+			validateArguments: (ctx, fixHints, arg1, arg2) => {
```
RandomByte, line 125:
```suggestion
			validateArguments: (ctx, fixHelper, arg1, arg2) => {
```

src/linter/ui5Types/fix/collections/sapUiCoreFixes.ts
```diff
@@ -112,9 +113,370 @@ t.declareModule("sap/ui/core/Core", [
...
+		// the resource bundle synchronously. When bAsync is true, the new API is not a replacement
+		// as it does not return a promise. In an await expression, it would be okay, but otherwise not.
+		// sLibrary must be a library.
+		t.method("getLibraryResourceBundle", callExpressionGeneratorFix({
+			moduleName: "sap/ui/core/Lib",
+			validateArguments: (ctx: {fallback: string}, fixHints, arg1, arg2, arg3) => {
+				// Handling fallback in the legacy API
+				if (!arg1 ||
+					(ts.isIdentifier(arg1) && arg1.text === "undefined") ||
+					(ts.isStringLiteralLike(arg1) && arg1.text === "undefined")) {
```
RandomByte, line 252: Checking for a string with the value `undefined` seems odd to me. Is this correct?

RandomByte: Got it, but thats the line above, right? `(ts.isIdentifier(arg1) && arg1.text === "undefined")`

I was wondering about  `(ts.isStringLiteralLike(arg1) && arg1.text === "undefined")`, which I think would only match a string like `"undefined"`

d3xter666: To be honest, I really don't recall this case 
I'll remove it for now

RandomByte (approved):

PR #700, opened 2025-06-13
refactor(ui5Types/utils): Enhance detection of return value usage

RandomByte:


test/lib/linter/ui5Types/utils.ts
```diff
@@ -380,6 +381,243 @@ test("getPropertyAssignmentsInObjectLiteralExpression - unsupported elements", (
...
+	t.true(isReturnValueUsed(conditionalExpression.condition));
+});
+
+test("isReturnValueUsed (negative): ConditionalExpression - whenTrue", (t) => {
+	const program = createProgram(`
+		condition ? method() : undefined;
+	`);
+	const expressionStatement = program.getSourceFile("test.ts")!.statements[0] as ts.ExpressionStatement;
+	const conditionalExpression = expressionStatement.expression as ts.ConditionalExpression;
+	t.false(isReturnValueUsed(conditionalExpression.whenTrue));
```
d3xter666, line 443: I'm wondering whether we should consider this as `true`?
Why would you have a ternary operator and not expect a valuie from it?

codeworrior: If the value is not used (void), then it doesn't matter.

Issue #657, opened 2025-05-13
Missing autofix related to jQuery.sap


flovogt:
# Missing autofix related to jQuery.sap

- [ ] `jQuery.sap.resources.isBundle`: Follow-Up of #521 
- [ ] `jQuery.sap.removeUrlWhitelist` Follow-Up of #525
- [ ] `jQuery.sap.delayedCall`, `jQuery.sap.clearDelayedCall`, `jQuery.sap.intervalCall` and `jQuery.sap.clearIntervalCall` Follow-Up of #590 (Fix #710)

d3xter666:
We'll skip the `jQuery.sap.removeUrlWhitelist` migration as the solution might bring more negatives than positives. 
The legacy solution is more elegant. Inside it uses a private method of `URLListValidator`.
However, if we want to migrate it, the solution would be:
- multiline
- will introduce local variables that we need to ensure are unique
- will need to find a way to keep the indentation of each row. And what about if `jQuery.sap.removeUrlWhitelist` is spread on multiple lines or has some special formatting with whitespaces inbetween....

Issue #521, opened 2025-02-07
Missing autofix for jQuery.sap.resources


flovogt:
# Missing autofix for jQuery.sap.resources

`jQuery.sap.resources` is deprecated. Therefore an autofix should be offered by UI5 linter.

## Deprecated Usage

````js
const bundle = jQuery.sap.resources({
  url: "resources/i18n.properties"
})
````

## Recommended Usage

````js
sap.ui.define([
  "sap/base/i18n/ResourceBundle"
], (ResourceBundle) => {
  const bundle = ResourceBundle.create({
    url: "resources/i18n.properties"
  });
}):
````

flovogt:
Will be implemented in CPOUI5FOUNDATION-990.

d3xter666:
`var isBundle = jQuery.sap.resources.isBundle(new sap.base.i18n.ResourceBundle());` shouls become
`var isBundle = new ResourceBundle() instanceof ResourceBundle;`

It's currently not migrated as the arguments of the call need to be replaced, too. We cannot fully support that with the initial release

PR #710, opened 2025-06-17
feat: Autofix for jQuery complex cases (multple iterations required)

d3xter666:
Resolves: https://github.com/SAP/ui5-linter/issues/657


src/linter/ui5Types/fix/collections/jqueryFixes.ts
```diff
@@ -714,6 +714,48 @@ t.declareModule("jQuery", [
...
+				}
+
+				return `setTimeout(${fnRepresentation}, ${timeout}${params ? ", ..." + params.trim() : ""})`;
+			},
+		})),
+		t.namespace("clearDelayedCall", callExpressionGeneratorFix({
+			generator: (_ctx, _, cbId) => {
+				return `clearTimeout(${cbId})`;
+			},
+		})),
```
RandomByte, line 737: No need for a generator fix

```suggestion
		t.namespace("clearDelayedCall", callExpressionFix({
			globalName: "clearTimeout",
		})),
```

src/linter/ui5Types/fix/collections/jqueryFixes.ts
```diff
@@ -714,6 +714,48 @@ t.declareModule("jQuery", [
...
+		t.namespace("intervalCall", callExpressionGeneratorFix({
+			validateArguments: (ctx: {fnNameNode: ts.Node}, _, _timeout, _objCtx, fnName) => {
```
RandomByte, line 739: You can request access to the global using the `globalName` property. This ensures that it doesn't conflict with an existing identifier in the source file. If it does, it will be prefixed automatically with `globalThis.`. The final property access string is provided in the second generator argument.

```suggestion
			globalName: "setInterval",
			validateArguments: (ctx: {fnNameNode: ts.Node}, _, _timeout, _objCtx, fnName) => {
```

src/linter/ui5Types/fix/collections/jqueryFixes.ts
```diff
@@ -116,6 +125,22 @@ t.declareModule("jQuery", [
...
+		// 		// TODO: Ensure aCurrentEntries is non conflicting with other variables
+		// 		// TODO: As this is a multiline solution, ensure to provide the leading space
+		// 		return `var aCurrentEntries = ${moduleIdentifier}.entries();\n` +
+		// 			`aCurrentEntries.splice(${iIndex}, 1);\n` +
+		// 			`${moduleIdentifier}.clear();\n` +
+		// 			`aCurrentEntries.forEach(function ({protocol, host, port, path}) {\n` +
+		// 			`\t${moduleIdentifier}.add(protocol, host, port, path);\n` +
+		// 			`});`;
+		// 	},
+		// })),
```
RandomByte, line 143: I'd propose to delete cases we don't plan to keep from our code base. We have the history here in the PR and I hereby also crosslink it to https://github.com/SAP/ui5-linter/issues/657#issuecomment-2983899399

It reduces readability of the file and the inactive code likely becomes invalid with the next refactoring anyways.

```suggestion
```

src/linter/ui5Types/fix/collections/jqueryFixes.ts
```diff
@@ -77,11 +77,20 @@ t.declareModule("jQuery", [
 				scope: FixScope.FirstChild,
 			})),
 		]),
+		t.namespace("resources", [
+			t.namespace("isBundle", callExpressionGeneratorFix({ // https://github.com/SAP/ui5-linter/issues/657
+				moduleName: "sap/base/i18n/ResourceBundle",
+				generator(_ctx, [moduleIdentifier], arg1) {
+					return `new ${moduleIdentifier}() instanceof ${arg1.trim()}`;
```
RandomByte, line 84: I think this needs to be swapped (compare to the example in https://github.com/SAP/ui5-linter/issues/521#issuecomment-2875641949 - which is _somewhat_ correct but easy to misunderstand).

My understanding is that:

```js
const myBundle = ResourceBundle.create({url : "i18n/messagebundle.properties"});
if (jQuery.sap.resources.isBundle(myBundle)) {
    // ...
}
```

Should be migrated to:

```js
const myBundle = ResourceBundle.create({url : "i18n/messagebundle.properties"});
if (myBundle instanceof ResourceBundle) {
    // ...
}
```

In fact, `new ResourceBundle()` as currently implemented here throws an error `resource URL 'undefined' has unknown type`

src/linter/ui5Types/fix/collections/jqueryFixes.ts
```diff
@@ -77,11 +77,20 @@ t.declareModule("jQuery", [
...
+				},
+			})),
+		],
 		// jQuery.sap.resources => ResourceBundle.create
-		t.namespace("resources", accessExpressionFix({ // https://github.com/SAP/ui5-linter/issues/521
+		accessExpressionFix({ // https://github.com/SAP/ui5-linter/issues/521
```
RandomByte, line 89: I think this should be a callExpressionFix (which then requires FullExpression scope I think).

Otherwise I fear we might incorrectly replace the property access in a larger property access, e.g. `jQuery.sap.resources.foo()`

```suggestion
		callExpressionFix({ // https://github.com/SAP/ui5-linter/issues/521
			scope: FixScope.FullExpression,
```

RandomByte (changes requested):

src/linter/ui5Types/fix/collections/jqueryFixes.ts
d3xter666: It's indeed a call expression. However, it should not be FullExpression scope, because the arguments are the same

RandomByte (approved):

PR #711, opened 2025-06-17
feat(autofix): Autofix basic property assignments in JS and XML

RandomByte:
Follow-ups:

* [x] Fix for `sap/ui/model/odata/v4/ODataModel#synchronizationMode` as outlined in JIRA
    * https://github.com/UI5/linter/pull/724
* [x] Validate XML before and after applying autofixes to ensure we don't break it (see https://github.com/SAP/ui5-linter/pull/711#pullrequestreview-2937890830)
    * https://github.com/UI5/linter/pull/725

JIRA: CPOUI5FOUNDATION-994


src/autofix/autofix.ts
```diff
@@ -264,18 +275,25 @@ export default async function ({
 		}
 	}
 
+	for (const resource of xmlResources) {
+		const resourcePath = resource.getPath();
+		const newContent = await applyFixesXml(resource, messages.get(resourcePath)!);
+		if (newContent) {
```
d3xter666, line 281: Could it be that the XML syntax breaks after the fix?
I was wondering whether we could do the same thing as with JS- check somehow for syntax errors after the fix and report them

src/autofix/generateChangesXml.ts
```diff
@@ -0,0 +1,115 @@
...
+import {Attribute, Position, SaxEventType, Tag} from "sax-wasm";
+import {parseXml} from "../utils/xmlParser.js";
+import Fix from "../linter/ui5Types/fix/Fix.js";
+import XmlEnabledFix from "../linter/ui5Types/fix/XmlEnabledFix.js";
+import {RawLintMessage} from "../linter/LinterContext.js";
+import {ChangeSet} from "./autofix.js";
+import {removeConflictingFixes} from "./utils.js";
+// import {getLogger} from "@ui5/logger";
+
+// const log = getLogger("linter:autofix:generateChangesXml");
```
d3xter666, line 11: leftover?

src/autofix/autofix.ts
RandomByte: Good point. sax-wasm might not be suitable for that purpose though. I'll check what options we have

RandomByte: If you don't mind, I would like to add XML validation in a separate PR. 

Implementing it might take longer, but since it doesn't fundamentally conflict with the work done here, it is probably easier to split it, also from a code review perspective.

d3xter666 (approved):

PR #714, opened 2025-06-18
refactor(autofix): Straighten out fix parameters

RandomByte:
Some parameters where available to generator-fixes even though those
don't use them.


RandomByte:
Re-triggered the failing e2e tests and now they are passing 🤷

PR #725, opened 2025-06-24
refactor(autofix): Validate XML before and after autofix

RandomByte:
Similar to the JS-autofix, we should validate XML for syntax errors
before and after applying fixes.

Add new dependency 'fast-xml-parser' which seems to have good reputation
and existing usage across SAP. It has only one transitive dependency and
overall reasonable footprint with ~600 KB install size. License is MIT.

Overall this seemed like a much easier solution than implementing home-grown validation based on the SAX parser. The new dependency is only (dynamically) imported if an autofix for an XML file is being applied.

JIRA: CPOUI5FOUNDATION-994


d3xter666 (approved):

src/autofix/autofix.ts
```diff
@@ -246,26 +271,41 @@ export default async function ({
...
+						if (d.line !== undefined && d.character !== undefined) {
+							// Prepend line and character
+							res = `${d.line + 1}:${d.character + 1} ${res}`;
+
+							// Fill contentWithMarkers array for debug logging
+							// Insert line below the finding and mark the character
+							const lineContent = contentWithMarkers[d.line + lineOffset];
+							// Count number of tabs until the character
+							const tabCount = lineContent.slice(0, d.character).split("\t").length - 1;
+							const leadingTabs = "\t".repeat(tabCount);
```
d3xter666, line 296: Could it be spaces or even worse- mixed style- tabs & spaces?
In such case of bad document formatting, the carret would be in the wrong position

PR #727, opened 2025-06-25
refactor(autofix): Add validation for SimpleForm#minWidth property fix

RandomByte:
Only if we detect that a layout other than ResponsiveLayout is used, we
can safely remove the minWidth property. If we detect that no layout is
set and therefore the default layout "ResponsiveGridLayout" is used,
we can also apply this fix.

If a binding is used to set the layout, no autofix should be applied.

JIRA: CPOUI5FOUNDATION-994


d3xter666 (approved):

d3xter666:
It seems that the cverage droped and that's why the checks fail. Otherwise LGTM

d3xter666 (approved):

PR #771, opened 2025-07-14
refactor(HTML): Encode every part of messageDetails

maxreichmann:
Since messageDetails without URLs may contain user input, they have to be encoded too.
Follow-up of [discussion](https://github.com/UI5/linter/pull/765/files/e35fe7e47d5811fa601ca94f47488c37bb6c5dc2#r2203932676) of #765.

JIRA: CPOUI5FOUNDATION-1061


matz3 (commented):
LGTM

PR #780, opened 2025-07-16
feat: Consider UI5 Manifest v2

d3xter666:
JIRA: CPOUI5FOUNDATION-1066


d3xter666:
Hey @KlattG,

Would you review the texts in:
- https://github.com/UI5/linter/pull/780/files#diff-d631fe82f8335589a701a32360cf291e6c3de7e2e48764273a337e41195271bc
- https://github.com/UI5/linter/pull/780/files#diff-09701c0232493655601b38f00fc2ebafebdf513a2aaf652ef5d440d1ea3ee04d

Cheers

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
 
 Checks whether the UI5 API is used correctly, for example, whether a formatter in a JavaScript/TypeScript binding declaration is of type `function`.
+
+## no-outdated-manifest-version
+
+For legacy free UI5, required manifest version is 2. This test checks the `_version` property in `manifest.json`.
+
+## no-removed-manifest-property
+
+In manifest version 2, some of the properties are removed and they do not match anymore to the new schema.
```
KlattG, line 133:
```suggestion
Manifest Version 2 lacks several properties of earlier versions, which do not match anymore to the new schema. They must be omitted.
```

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+## no-renamed-manifest-property
+
+In manifest version 2, some of the properties are renamed, so the new names must be adjusted to match the schema.
```
KlattG, line 137:
```suggestion
Manifest Version 2 renames several properties of earlier versions. Their names must be adjusted to match the new schema.
```

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+## no-legacy-ui5-version-in-manifest-2
+
+Check for legacy UI5 versions in `minUI5Version` property in `manifest.json`
```
KlattG, line 141:
```suggestion
Checks the `minUI5Version` property in `manifest.json` for legacy UI5 versions, which must not be used.
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
 			"See {@link topic:28fcd55b04654977b63dacbee0552712 Best Practices for Developers}",
 	},
 
+	[MESSAGE.NO_OUTDATED_MANIFEST_VERSION]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-outdated-manifest-version"],
+
+		message: () => "manifest.json must be migrated to version 2",
+		details: () =>
+			"Set _version to >= 2.0.0 and follow the migration guide {@link topic:be0cf40f61184b358b5faedaec98b2da " +
```
KlattG, line 702:
```suggestion
			"Set _version to >= 2.0.0 and follow the migration guide {@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 " +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			"Manifest (Descriptor for Applications, Components, and Libraries)}",
```
KlattG, line 703:
```suggestion
			"Manifest Version 2}",
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+	},
+
+	[MESSAGE.NO_LEGACY_UI5_VERSION_IN_MANIFEST_2]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-legacy-ui5-version-in-manifest-2"],
+
+		message: () => "Use legacy free version of UI5 in manifest.json.",
```
KlattG, line 710:
```suggestion
		message: () => "Use a legacy-free UI5 version in manifest.json.",
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+		details: () =>
+			"Set minUI5Version to >= 1.136.0 or higher and follow the migration " +
+			"guide {@link topic:be0cf40f61184b358b5faedaec98b2da " +
```
KlattG, line 713:
```suggestion
			"guide {@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 " +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			"Manifest (Descriptor for Applications, Components, and Libraries)}",
```
KlattG, line 714:
```suggestion
			"Manifest Version 2}",
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+	},
+
+	[MESSAGE.NO_REMOVED_MANIFEST_PROPERTY]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-removed-manifest-property"],
+
+		message: ({propName}: {propName: string}) => `Property '${propName}' has been removed in manifest ` +
```
KlattG, line 721:
```suggestion
		message: ({propName}: {propName: string}) => `Property '${propName}' has been removed in Manifest ` +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			`version 2 and must no longer be provided`,
```
KlattG, line 722:
```suggestion
			`Version 2 and must no longer be provided`,
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			`guide {@link topic:be0cf40f61184b358b5faedaec98b2da Manifest (Descriptor for Applications, ` +
+			`Components, and Libraries)}`,
+	},
+
+	[MESSAGE.NO_RENAMED_MANIFEST_PROPERTY]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-renamed-manifest-property"],
+
+		message: ({propName, newName}: {propName: string; newName: string}) =>
+			`Property '${propName}' has been renamed to '${newName}' in manifest version 2`,
```
KlattG, line 733:
```suggestion
			`Property '${propName}' has been renamed to '${newName}' in Manifest Version 2`,
```

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+## no-incorrect-manifest-property-value
+
+Checks whether property `type` is explicitly set to `View` in manifest version 2 when `sap.ui5/routing/config/type` is not omitted or set to `View`.
```
KlattG, line 145:
```suggestion
In Manifest Version 2, checks whether the `type` property is explicitly set to `View` when `sap.ui5/routing/config/type` is not omitted or set to `View`.
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+	},
+
+	[MESSAGE.NO_INCORRECT_MANIFEST_PROPERTY_VALUE]: {
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-incorrect-manifest-property-value"],
+
+		message: ({propName, value}: {propName: string; value: string}) =>
+			`Check if '${propName}' is set to '${value}'`,
+		details: () =>
+			`Property 'type' should be explicitly set to 'View' in manifest version 2 ` +
```
KlattG, line 747:
```suggestion
			`Property 'type' must be explicitly set to 'View' in Manifest Version 2 ` +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			`if 'sap.ui5/routing/config/type' is not omitted or set to 'View'. ` +
+			`{@link topic:be0cf40f61184b358b5faedaec98b2da Manifest (Descriptor for Applications, ` +
```
KlattG, line 749:
```suggestion
			`See {@link topic:be0cf40f61184b358b5faedaec98b2da Manifest (Descriptor for Applications, ` +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+	},
+
+	[MESSAGE.NO_REMOVED_MANIFEST_PROPERTY]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-removed-manifest-property"],
+
+		message: ({propName}: {propName: string}) => `Property '${propName}' has been removed in manifest ` +
+			`version 2 and must no longer be provided`,
+		details: ({propName}: {propName: string}) => `Remove property '${propName}' and follow the migration ` +
+			`guide {@link topic:be0cf40f61184b358b5faedaec98b2da Manifest (Descriptor for Applications, ` +
```
KlattG, line 724:
```suggestion
			`guide {@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 Manifest  ` +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			`Components, and Libraries)}`,
```
KlattG, line 725:
```suggestion
			`Version 2}`,
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+
+	[MESSAGE.NO_RENAMED_MANIFEST_PROPERTY]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-renamed-manifest-property"],
+
+		message: ({propName, newName}: {propName: string; newName: string}) =>
+			`Property '${propName}' has been renamed to '${newName}' in manifest version 2`,
+		details: ({propName, newName}: {propName: string; newName: string}) =>
+			`Rename property '${propName}' to '${newName}' and follow the migration guide ` +
+			`{@link topic:be0cf40f61184b358b5faedaec98b2da Manifest (Descriptor for Applications, ` +
```
KlattG, line 736:
```suggestion
			`{@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 Manifest ` +
```

src/linter/messages.ts
```diff
@@ -683,4 +693,61 @@ export const MESSAGE_INFO = {
...
+			`Components, and Libraries)}`,
```
KlattG, line 737:
```suggestion
			`Version 2}`,
```

KlattG (dismissed):
LGTM

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+## no-removed-manifest-property
+
+Manifest Version 2 lacks several properties of earlier versions, which do not match anymore to the new schema. They must be omitted.
+
+## no-renamed-manifest-property
+
+Manifest Version 2 renames several properties of earlier versions. Their names must be adjusted to match the new schema.
+
+## no-legacy-ui5-version-in-manifest-2
```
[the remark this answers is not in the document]
d3xter666: According to the BLI, it should always recommend usage of non legacy versions. So, it should be checked independent of the _version property. I'll update the rule ID

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +57,35 @@ export default class ManifestLinter {
...
+		if ("minUI5Version" in manifest) {
+			let availableVersions: string[] = [];
+
+			if (Array.isArray(manifest.minUI5Version)) {
+				availableVersions = manifest.minUI5Version as string[];
+			} else if (typeof manifest.minUI5Version === "string") {
+				availableVersions.push(manifest.minUI5Version);
+			}
+
+			// Check if any version is below 1.36
```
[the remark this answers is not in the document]
d3xter666: Done

test/fixtures/linter/rules/NoDeprecatedApi/manifest.json
```diff
@@ -1,5 +1,7 @@
 {
 	"_version": "1.12.0",
+	
+	"minUI5Version": ["1.119.0"],
```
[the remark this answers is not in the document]
d3xter666: Removed

docs/Rules.md
```diff
@@ -118,3 +123,23 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+Checks the `minUI5Version` property in `manifest.json` for legacy UI5 versions, which must not be used.
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/messages.ts
```diff
@@ -683,4 +693,62 @@ export const MESSAGE_INFO = {
...
+			"Manifest Version 2}",
+	},
+
+	[MESSAGE.NO_LEGACY_UI5_VERSION_IN_MANIFEST_2]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["no-legacy-ui5-version-in-manifest-2"],
+
+		message: () => "Use a legacy-free UI5 version in manifest.json.",
+		details: () =>
+			"Set minUI5Version to >= 1.136.0 or higher and follow the migration guide " +
```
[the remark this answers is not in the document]
d3xter666: Thanks! Removed the comparison sign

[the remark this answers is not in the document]
matz3: With that it's not consistent with other new rules such as the one above that says `Set _version to >= 2.0.0`. I think we should be consistent here.

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +57,35 @@ export default class ManifestLinter {
...
+
+		if (manifest?._version?.startsWith("2.")) {
+			this.#analyzeManifest_1(manifest, true);
+		} else {
+			this.#analyzeManifest_1(manifest);
+			this.#reporter?.addMessage(MESSAGE.NO_OUTDATED_MANIFEST_VERSION, {} as never, "/_version");
+		}
+	}
+
+	#analyzeManifest_1(manifest: SAPJSONSchemaForWebApplicationManifestFile, isManifest2 = false) {
```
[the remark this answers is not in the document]
d3xter666: Hopefully `validatePropertiesForManifestVersion` would make more sense. My initial idea was to have a separate validators for manifest version 1 and version 2, but it was not that convenient

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +57,35 @@ export default class ManifestLinter {
...
+			if (Array.isArray(manifest.minUI5Version)) {
+				availableVersions = manifest.minUI5Version as string[];
+			} else if (typeof manifest.minUI5Version === "string") {
+				availableVersions.push(manifest.minUI5Version);
+			}
+
+			// Check if any version is below 1.36
+			const isBellow136 = availableVersions.some((version) => {
+				const [major, minor] = version.split(".").map(Number);
+				return major === 1 && minor < 136;
```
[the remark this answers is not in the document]
d3xter666: Added semver

src/linter/messages.ts
```diff
@@ -683,4 +693,62 @@ export const MESSAGE_INFO = {
...
+
+		message: ({propName, newName}: {propName: string; newName: string}) =>
+			`Property '${propName}' has been renamed to '${newName}' in Manifest Version 2`,
+		details: ({propName, newName}: {propName: string; newName: string}) =>
+			`Rename property '${propName}' to '${newName}' and follow the migration guide ` +
+			`{@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 ` +
+			`Manifest Version 2}`,
+	},
+
+	[MESSAGE.NO_INCORRECT_MANIFEST_PROPERTY_VALUE]: {
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/messages.ts
```diff
@@ -683,4 +693,62 @@ export const MESSAGE_INFO = {
...
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-incorrect-manifest-property-value"],
+
+		message: ({propName, value}: {propName: string; value: string}) =>
+			`Check if '${propName}' is set to '${value}'`,
+		details: () =>
+			`Property 'type' must be explicitly set to 'View' in Manifest Version 2 ` +
```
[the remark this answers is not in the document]
d3xter666: I have adjusted it to be more specific! Thanks!

src/linter/messages.ts
```diff
@@ -683,4 +693,62 @@ export const MESSAGE_INFO = {
...
+			`{@link topic:be0cf40f61184b358b5faedaec98b2da#loiobe0cf40f61184b358b5faedaec98b2da/section_manifest2 ` +
+			`Manifest Version 2}`,
+	},
+
+	[MESSAGE.NO_INCORRECT_MANIFEST_PROPERTY_VALUE]: {
+		severity: LintMessageSeverity.Warning,
+		ruleId: RULES["no-incorrect-manifest-property-value"],
+
+		message: ({propName, value}: {propName: string; value: string}) =>
+			`Check if '${propName}' is set to '${value}'`,
```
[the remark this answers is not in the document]
d3xter666: That's the description of this finding: 
>If /sap.ui5/routing/config/type does not contain the value „View“ e.g. "Component", every target which is a view has to set the „type“ Property to value „View“ manually, or as an optimization „/sap.ui5/routing/config/type“ could be removed and only for the non view targets could the corresponding „type“ Property be set

We should either recommend or do not report. I would appreciate any other suggestions on how to improve the situation.

src/linter/messages.ts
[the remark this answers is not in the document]
d3xter666: Now should be better

src/linter/messages.ts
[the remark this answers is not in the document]
d3xter666: Done

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,35 +58,35 @@ export default class ManifestLinter {
...
-				return major === 1 && minor < 136;
+			// Check if any version is below 1.136
+			const isBelow136 = availableVersions.some((version) => {
+				return semver.lt(version, "1.136.0");
 			});
 
-			if (isBellow136) {
-				this.#reporter?.addMessage(MESSAGE.NO_LEGACY_UI5_VERSION_IN_MANIFEST_2, {} as never, "/minUI5Version");
+			if (isBelow136) {
+				this.#reporter?.addMessage(MESSAGE.NO_LEGACY_UI5_VERSION_IN_MANIFEST, {} as never,
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/linter/messages.ts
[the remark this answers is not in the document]
codeworrior: Not sure if I got it right. If the default (`sap.ui5/routing/config/type`) is given and not "View", then this means any target without an own "type" property will be of that default type. None can be of type "View", except if the type is given in the target itself.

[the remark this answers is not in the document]
d3xter666: This thread relates also to this comment: https://github.com/UI5/linter/pull/780#discussion_r2218710109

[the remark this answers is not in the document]
d3xter666: > Not sure if I got it right. If the default (`sap.ui5/routing/config/type`) is given and not "View", then this means any target without an own "type" property will be of that default type. None can be of type "View", except if the type is given in the target itself.

Exactly!

And my question is:
- Shall we check if someone has missed `type: "View"` by accident in that case?
- If "yes", then how can we reliably check if a target is a "View"?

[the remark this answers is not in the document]
codeworrior: No, I wouldn't check this. The app would already have failed with a runtime error whenever such a misconfigured target was reached.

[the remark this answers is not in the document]
d3xter666: Such a relief! 😅 
Thank you @codeworrior !

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -89,16 +120,49 @@ export default class ManifestLinter {
 		}
 
 		// Detect deprecated view type in routing.config:
-		if (routing?.config && routing.config.viewType && deprecatedViewTypes.includes(routing.config.viewType)) {
+		const viewType = routing?.config?.viewType as string ?? routing?.config?.type;
```
matz3, line 123: Isn't `viewType` and `type` something completely different?
https://github.com/SAP/ui5-manifest/blob/4e79c6f80d21cce52c52394877b61384c53d3ed4/schema.json#L8206-L8216
vs.
https://github.com/SAP/ui5-manifest/blob/4e79c6f80d21cce52c52394877b61384c53d3ed4/schema.json#L8157-L8164

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -89,16 +120,49 @@ export default class ManifestLinter {
...
+			(routing?.config?.viewType ? "/sap.ui5/routing/config/viewType" : "/sap.ui5/routing/config/type"));
 		}
 
 		// Detect deprecations in routing.targets:
 		const targets = routing?.targets;
+		const oldToNewTargetPropsMap = {
+			viewName: "name",
+			viewId: "id",
+			viewLevel: "level",
+			viewPath: "path",
```
matz3, line 137: Those properties may also be present in `routing.config`, where they should also be checked.

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +58,36 @@ export default class ManifestLinter {
...
+				this.#reporter?.addMessage(MESSAGE.NO_LEGACY_UI5_VERSION_IN_MANIFEST,
+					"/sap.ui5/dependencies/minUI5Version");
+			}
+		}
+
+		if (manifest?._version?.startsWith("2.")) {
+			this.#validatePropertiesForManifestVersion(manifest, true);
+		} else {
+			this.#validatePropertiesForManifestVersion(manifest);
+			this.#reporter?.addMessage(MESSAGE.NO_OUTDATED_MANIFEST_VERSION, "/_version");
```
matz3, line 86: Note that the manifest version 2 is currently not supported by the UI5 Language Assistant (see https://github.com/SAP/ui5-language-assistant/issues/748). We have to keep that in mind as people using the extension will see an error when changing the version to `2.0.0`.

src/linter/manifestJson/ManifestLinter.ts
d3xter666: Yes and No 😅 

`type` used to be what `viewType` now represents. With the introduction of `viewType`, the `type` property started to have the values `View` and `Component`. But there are many places where `type` is still used to handle `XML`, `JS`, etc.

That's why we check if `viewType` is present first and then fallback to `type` here

src/linter/manifestJson/ManifestLinter.ts
d3xter666: Thanks!

src/linter/manifestJson/ManifestLinter.ts
codeworrior: We had a meeting on July 7 where this was discussed with the owners of ui5-language-assistant and they agreed on an implementation plan, using `_version` and URLs like https://raw.githubusercontent.com/SAP/ui5-manifest/refs/tags/v2.0.0/schema.json to access the schema.

src/linter/manifestJson/ManifestLinter.ts
matz3: I'm not aware of such a change. The only place where I can see `type` is used for a view type is within the `sap.ui5/rootView`. Can you link to the relevant part in the schema?

codeworrior: As commented [here](https://github.com/UI5/linter/pull/780#discussion_r2245865507), I could not find any evidence that there ever was a renaming of those properties in the "sap.ui5"/"routing".

Even in 1.38, the `viewType` was already named `viewType`:
https://ui5.sap.com/1.38.64/#docs/guide/902313063d6f45aeaa3388cc4c13c34e.html

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -89,16 +120,59 @@ export default class ManifestLinter {
...
+								newName: newProp,
+							}, `/sap.ui5/routing/targets/${key}/${oldProp}`);
+						}
+					}
+
+					if (configType === undefined && target?.type === undefined) {
+						// Type must be defined somewhere, either in routing.config or in every target.type
+						this.#reporter?.addMessage(MESSAGE.NO_MISSING_MANIFEST_CONFIGURATION, {
+							propertyPath: `/sap.ui5/routing/targets/${key}/type`,
+						}, `/sap.ui5/routing/targets/${key}/type`);
```
matz3, line 167: This message is always reported at 0:0 or 1:1, as the location does not exist. Should we maybe rather use the target object as location?

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -89,16 +120,59 @@ export default class ManifestLinter {
...
+						if (target[oldProp] && !target[newProp]) {
+							this.#reporter?.addMessage(MESSAGE.NO_RENAMED_MANIFEST_PROPERTY, {
+								propName: oldProp,
+								newName: newProp,
+							}, `/sap.ui5/routing/targets/${key}/${oldProp}`);
+						}
+					}
+
+					if (configType === undefined && target?.type === undefined) {
+						// Type must be defined somewhere, either in routing.config or in every target.type
```
matz3, line 164: This comment explains what to do, but I think that is missing in the message. When there are several targets, the current behavior is that for EVERY target you get a message asking to add the property to the target. But usually, IMHO it's easier and better to add the `type: "View"` to the general config, and use `type: "Component"` on targets where a component instead of a view is targeted.

cc @stopcoder

src/linter/manifestJson/ManifestLinter.ts
d3xter666: Thanks! Fixed it!

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -89,16 +120,57 @@ export default class ManifestLinter {
...
+							this.#reporter?.addMessage(MESSAGE.NO_RENAMED_MANIFEST_PROPERTY, {
+								propName: oldProp,
+								newName: newProp,
+							}, `/sap.ui5/routing/targets/${key}/${oldProp}`);
+						}
+					}
+
+					if (configType === "View" && target?.type === "View") {
+						// When routing.config.type is "View" setting target.type "View" is redundant,
+						this.#reporter?.addMessage(MESSAGE.REDUNDANT_VIEW_CONFIG_PROPERTY, {
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +58,36 @@ export default class ManifestLinter {
 	}
 
 	#analyzeManifest(manifest: SAPJSONSchemaForWebApplicationManifestFile) {
+		if (manifest?.["sap.ui5"]?.dependencies?.minUI5Version) {
+			let availableVersions: string[] = [];
+
+			if (Array.isArray(manifest?.["sap.ui5"]?.dependencies?.minUI5Version)) {
+				availableVersions = manifest?.["sap.ui5"]?.dependencies?.minUI5Version;
```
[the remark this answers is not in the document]
d3xter666: Thanks!

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -117,10 +189,37 @@ export default class ManifestLinter {
...
+						this.#reporter?.addMessage(MESSAGE.NO_MISSING_MANIFEST_TARGET_TYPE, {
+							propertyPath: `/sap.ui5/routing/targets/${key}/type`,
+						}, `/sap.ui5/routing/targets/${key}`);
+					});
+				} else {
+					this.#reporter?.addMessage(MESSAGE.NO_MISSING_MANIFEST_TARGET_TYPE, {
+						propertyPath: `/sap.ui5/routing/config/type`,
+					}, `/sap.ui5/routing/config`);
+				}
+			}
```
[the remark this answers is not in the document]
matz3: Well, now I'm a bit confused. We already had those checks implemented a while ago but removed them because a migration guide in our documentation was missing (see https://github.com/UI5/linter/pull/356).

The properties like `viewName`, `viewLevel`, etc. are now only checked in manifest version 2, but the `type` is checked in all versions. This doesn't seem right to me as both are connected. I might be wrong here and would need to check the docs and code again, but my understanding is that if you use `type: "View"`, then you MUST provide the properties as `name`, `level`, etc. But if you don't use the `type`, which was introduced later, you still have to use the prefixed properties (again, I might be wrong here). But those prefixed properties are deprecated and won't work in a legacy-free UI5 version, so we should report them.
It feels like an easy check at first glance but becomes rather complex when thinking about all possible cases. I also think that better unit testing would help to cover all cases. Currently we don't have a lot of different cases that we test, although most of the code seems to be covered by tests.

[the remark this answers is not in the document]
d3xter666: Alright, I might have missed that!

Initially, my understanding was that `type` was introduced with Manifest 2, the renaming of the props, too. But, obviously my conclusions were wrong...
Given that `type` is available since 1.38.x (and maybe even earlier), then it should be Manifest agnostic, IMO.
However, when we include the renamed properties, give your input, the check should become a lot more complex...

I'd need to check the code for that...

@devtomtom can you give us your oppinion here, please?

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -57,6 +58,36 @@ export default class ManifestLinter {
 	}
 
 	#analyzeManifest(manifest: SAPJSONSchemaForWebApplicationManifestFile) {
```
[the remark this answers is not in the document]
matz3: No, I think we should currently only check files named `manifest.json`. If we are checking also `manifest.appdescr_variant` files, that should be a separate check as the structure differs.

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -42,6 +43,11 @@ export default class ManifestLinter {
 
 	// eslint-disable-next-line @typescript-eslint/require-await
 	async lint() {
+		// Only lint manifest.json files and skip manifest.appdescr_variant, for example
+		if (!this.#resourcePath.includes("manifest.json")) {
```
matz3, line 47: Shouldn't we rather do this check before even creating an instance and calling `lint`?
See `src/linter/manifestJson/linter.ts`. I don't remember a reason why `manifest.appdescr_variant` has been included there. Maybe @RandomByte knows.

codeworrior: In the context of UI5 Tooling, we learned earlier the hard way, that `manifest.appdesc_variant` is also an indicator for a "project of interest", much like `manifest.json` is. 

I guess this led to the (wrong) conclusion that it should be handled the same way as a `manifest.json` in the linter. But in fact, it contains Flex "descriptor changes", not the usual manifest content. There could be something to check there as well (e.g. there's a descriptor change to add a model to the original manifest.json and the change in theory could contain the deprecated setting synchronizationMode for an OData V4 model), but I guess that's out of scope for this PR.

If interested and if you have access to our internal Wiki, you can view "Available+Descriptor+Change+Merger+Implementations" and search for `appdescr_ui5_addNewModel` for details about the change types.

src/linter/manifestJson/ManifestLinter.ts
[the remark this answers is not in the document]
d3xter666: Thanks! Then, I will remove the Glob pattern in manifest's `linter.ts`, so that it matches only `manifest.json` files. If we need the `manifest.appdescr_variant` later, then we can create a separate module or check for it.

src/linter/manifestJson/ManifestLinter.ts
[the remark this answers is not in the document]
d3xter666: The `routing/config/async` has a new default value for Manifest 2 (according to CPOUI5FOUNDATION-935) which makes me doubt wether we need to hanlde it seperately here.

Basically, if we enforce removing the `routing/config/async` flag in CPOUI5FOUNDATION-935, then the whole complex implementation here would become redundant.

What's your oppinion on that?

src/linter/manifestJson/ManifestLinter.ts
RandomByte: It seems like this was added in https://github.com/UI5/linter/pull/69/files#diff-d4d8a4a1e91835bf84a4cd0f2b024ea395b75a6b872e562dc9452ec03cd4e55e

I sadly can't recall why I added it there. I thought we had discussed it somewhere, but couldn't find any notes. Seems reasonable to remove it.

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -101,10 +145,40 @@ export default class ManifestLinter {
 					}, `${pathToViewObject}/viewType`);
 				}
 			}
+
+			// "type" must be defined somewhere, either in routing.config or in every target.type
+			if (configType === undefined && targetsType.length) {
```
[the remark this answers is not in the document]
d3xter666: Done

src/linter/messages.ts
```diff
@@ -144,6 +151,19 @@ export const MESSAGE_INFO = {
 			`{@link sap.ui.core.IAsyncContentCreation sap.ui.core.IAsyncContentCreation}`,
 	},
 
+	[MESSAGE.MANIFEST_ASYNC_FALSE_ERROR]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["async-component-flags"],
+
+		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
+			`The 'async: false' flag at '${asyncFlagLocation}' is not recommended`,
```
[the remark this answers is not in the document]
d3xter666: It depends on the schema version- for v2 it's forbidden as it's not in the schema and for v1 is valid, but is antipattern as it makes sync requests.
I wanted it to be more generic for v1 and v2 and maybe this is the core issue here.

src/linter/messages.ts
```diff
@@ -144,6 +151,19 @@ export const MESSAGE_INFO = {
...
+		details: () =>
+			`Setting 'async' to 'false' prevents asynchronous loading and is not recommended. ` +
```
KlattG, line 161: @d3xter666 see the discussion above. So perhaps for Manifest 1.x write 
`The 'async' parameter must not be set to 'false'.  ` +
and for Manifest 2.x
`The 'async' parameter must not be provided.  ` +

KlattG (changes requested):
I've made some suggestions but am not quite certain how to treat the async parameter. Should it be still allowed in Version 1 when set to true and not be allowed in Version 2 at all? Or should we rather get rid of it completely? Up to you to decide.

src/linter/messages.ts
KlattG: I think you'll have to distinguish between the versions then, like this:

**Manifest Version 1:**
		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
			`The 'async' property at '${asyncFlagLocation}' must be set to 'true'. `,
		details: () =>
           `Setting 'async' to 'false' prevents asynchronous loading and is not recommended. ` +
			`Set 'async: true' and implement the sap.ui.core.IAsyncContentCreation interface ` +
			`for proper asynchronous loading support. ` +
			`{@link topic:676b636446c94eada183b1218a824717 Use Asynchronous Loading}`,
	},

**Manifest Version 2:**
		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
			`The 'async' property at '${asyncFlagLocation}' must not be specified. `,
		details: () =>
			`Remove the 'async' flag and implement the sap.ui.core.IAsyncContentCreation interface ` +
			`for proper asynchronous loading support. ` +
			`{@link topic:b430345887f1419fba50320b57c1bdf9 Methods Controlling the Initial Instantiation}`,

src/linter/messages.ts
```diff
@@ -144,6 +151,19 @@ export const MESSAGE_INFO = {
 			`{@link sap.ui.core.IAsyncContentCreation sap.ui.core.IAsyncContentCreation}`,
 	},
 
+	[MESSAGE.MANIFEST_ASYNC_FALSE_ERROR]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["async-component-flags"],
+
+		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
+			`The 'async' property at '${asyncFlagLocation}' must be set to 'true'.`,
```
matz3, line 159: Here, setting `async: true` is the first recommendation to solve the issue. The message COMPONENT_MISSING_ASYNC_INTERFACE mentions `Alternatively, set the 'async' flag to 'true'`, so there it's the opposite. Is this intended?

For example, when `_version` is `1.x.x`, and `IAsyncContentCreation` is used:
1. Defining `routing/config/async=false` causes the message: `The 'async' property at '/sap.ui5/routing/config/async' must be set to 'true'`
2. Setting it to `true` then causes the message: `Component implements the sap.ui.core.IAsyncContentCreation interface. The redundant 'async' flag at '/sap.ui5/routing/config/async' should be removed from the component manifest`

Apart from that, the messages seem okay to me, but the duplicated checks in asyncComponentFlags and ManifestLinter makes it very complex to check/review and it's easy that some constellation is missed or behaves inconsistent. I have not checked all combinations.

d3xter666: The `COMPONENT_MISSING_ASYNC_INTERFACE` is triggered when the async interface is not implemented and also `async` flags are either not set or `false` (this is the manifest version 1 case).

However, if we have the `sap.ui.core.IAsyncContentCreation` implemented, but there's still `async: false` property set, then we need to point it. And that's exactly the reson behind this message. The `COMPONENT_MISSING_ASYNC_INTERFACE` would not be correct here as there is already the async interface implemented.

matz3: Yes, and I think the recent change to not add a message when IAsyncContentCreation is not implemented makes sense as there is already a message that suggests the usage of it.

Still, the scenario I described above still applies and it's a bit unfortunate for users to be first advised to change the property to `true`, and in the next run you are advised to remove it completely. Can't we recommend the removal right from the beginning? At the end, it's not wrong, but a bit unexpected.

matz3: I would have expected that, if the `IAsyncContentCreation` interface is used, and `async: false` is defined, we recommend to remove the property. This is the result of the scenario I described above, but it currently needs two runs to get to it.

KlattG: Hmm, I think we might need to consider all possible combinations and have different messages if needed.

**Manifest v2** is easy, because there is no async property anymore - if one is found, it **must** be removed. If no IAsyncContentCreation interface is used, we should recommend it.

For **Manifest v1**, I see 4 cases: 
1. No IAsyncContentCreation interface and async: false --> recommend usage of IAsync and removal of async
2. No IAsyncContentCreation interface and async: true --> recommend usage of IAsync and removal of async
3. IAsyncContentCreation interface and async: false --> recommend removal of async
4. IAsyncContentCreation interface and async: true --> recommend removal of async

If I understood @d3xter666  correctly, some of these recommendations might break an existing app, however. In this case, we need to add a warning to our message, right?

codeworrior: The critical cases where the app has to check the outcome are 1) and to some degree 2).

src/linter/messages.ts
```diff
@@ -144,6 +151,22 @@ export const MESSAGE_INFO = {
...
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["async-component-flags"],
+
+		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
+			`The 'async' property at '${asyncFlagLocation}' must be removed`,
+		details: ({hasAsyncInterface}: {hasAsyncInterface?: boolean}) =>
+			`Setting 'async' to 'false' prevents asynchronous loading and is not recommended. Remove the 'async' flag` +
+			(hasAsyncInterface ?
+				". " :
+				` and verify if the sap.ui.core.IAsyncContentCreation ` +
```
KlattG, line 164:
```suggestion
				` and verify that the sap.ui.core.IAsyncContentCreation ` +
```

src/linter/manifestJson/ManifestLinter.ts
d3xter666: This one has been resolved by removing the `manifest.appdesc_variant` from the glob pattern. Now we check only `manifest.json`

KlattG (approved):
LGTM

src/linter/messages.ts
```diff
@@ -144,6 +151,22 @@ export const MESSAGE_INFO = {
...
 	},
 
+	[MESSAGE.MANIFEST_ASYNC_FALSE_ERROR]: {
+		severity: LintMessageSeverity.Error,
+		ruleId: RULES["async-component-flags"],
+
+		message: ({asyncFlagLocation}: {asyncFlagLocation: string}) =>
+			`The 'async' property at '${asyncFlagLocation}' must be removed`,
+		details: ({hasAsyncInterface}: {hasAsyncInterface?: boolean}) =>
+			`Setting 'async' to 'false' prevents asynchronous loading and is not recommended. Remove the 'async' flag` +
```
codeworrior, line 161: At least in the context of the new checks, "not recommended" is too weak. it's not supported.

src/linter/ui5Types/asyncComponentFlags.ts
```diff
@@ -346,9 +354,63 @@ function reportUiComponentResults({
...
+
+	if (isManifestV2) {
+		// Only report errors when no external manifest.json exists to avoid duplication with ManifestLinter
+		if (manifestContent) {
+			// Avoid duplicate checks
+			return;
+		}
+		// Inline manifest definition (within Component.js)
+		// Manifest v2 logic:
+		// - Do not enforce IAsyncContentCreation (if present, do nothing)
```
codeworrior, line 371: I don't see that the existence of the interface would have an impact on the below checks. To me, the comment seems outdated.

src/linter/manifestJson/ManifestLinter.ts
```diff
@@ -167,4 +204,45 @@ export default class ManifestLinter {
...
+				this.#reporter?.addMessage(MESSAGE.NO_REMOVED_MANIFEST_PROPERTY, {
+					propName: "/sap.ui5/routing/config/async",
+				}, "/sap.ui5/routing/config/async");
+			} else if (routing.config.async === false) {
+				// Error: async: false is not supported
+				this.#reporter?.addMessage(MESSAGE.MANIFEST_ASYNC_FALSE_ERROR, {
+					asyncFlagLocation: "/sap.ui5/routing/config/async",
+				}, "/sap.ui5/routing/config/async");
+			}
+			// Note: Same as above - IAsyncContentCreation validation is handled elsewhere
```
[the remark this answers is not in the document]
d3xter666: Thanks!

I believe it's a lefteover after some refactorings.
I have removed it now

docs/Rules.md
```diff
@@ -118,3 +121,15 @@ Checks whether the declaration of UI5 classes is correct. This rule only applies
...
+
+Legacy-free UI5 requires Manifest Version 2. This test checks the `_version` property in `manifest.json`.
+
+## no-removed-manifest-property
+
+Manifest Version 2 lacks several properties of earlier versions, which do not match anymore to the new schema. They must be omitted.
+
+## no-legacy-ui5-version-in-manifest
+
+Checks the `sap.ui5/dependencies/minUI5Version` property in `manifest.json` for legacy UI5 versions, which must not be used.
```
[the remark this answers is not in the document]
d3xter666: Thanks! 
I have updated the documentation!

Hopefuly, it's clearer now

src/linter/manifestJson/ManifestLinter.ts
codeworrior: Meanwhile, I got a reference to the corresponding action item on their side: 
Issue \#34962 in the (SAP internal) ux-engineering/tools-suite repo.

PR #826, opened 2025-08-21
docs: Autofix restrictions for jQuery plugins / Third-party modules

matz3:


codeworrior (approved):
The reason for all 3 gaps is the lack of type information / resp. api.json information for them. As this document is intended for end-users of the linter, it might not be worth mentioning it.

PR #842, opened 2025-08-29
fix: Ensure deterministic order of processing source files

matz3:
In some rare cases, the non-deterministic processing within the TypeScript compiler could cause different outcomes of the linting. This may have happened when the same global type has been defined in multiple files.

### DIFF
--- a/test/lib/linter/ui5Types/host.ts
+++ b/test/lib/linter/ui5Types/host.ts
@@ -187,5 +187,5 @@ test("createVirtualLanguageServiceHost: getScriptFileNames returns sorted files"
 	];
 
 	// Check that our test files are in the correct sorted order
-	t.deepEqual(scriptFileNames, expectedOrder);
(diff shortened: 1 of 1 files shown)


test/lib/linter/ui5Types/host.ts
```diff
@@ -156,3 +156,36 @@ test("createVirtualLanguageServiceHost: Minimal project with sap/m/Button import
...
+
+	// Verify the files are actually sorted
+	const expectedOrder = [
+		"/resources/test/a.js",
+		"/resources/test/m.js",
+		"/resources/test/z.js",
+	];
+
+	// Check that our test files are in the correct sorted order
+	t.deepEqual(scriptFileNames, expectedOrder);
```
maxreichmann, line 190: Nitpick: 
```suggestion
	t.deepEqual(scriptFileNames, expectedOrder, "Resources should be sorted by filename");
```

maxreichmann (approved):
Just one nitpick.
Apart from that, LGTM!

PR #870, opened 2025-09-17
fix: Release the SharedLanguageService even at exception

d3xter666:
The SharedLanguageService provides information to the TS compiler and does some additional tasks like transforming XML to JS.
When acquired, it must be released, but in case of an exception during the XML transformation, it was not released.


codeworrior (approved):
LGTM

PR #882, opened 2025-09-29
ci(github-actions): Ignore Coveralls errors

maxreichmann:
This will add `continue-on-error` to the Coveralls report upload step in the job "CI / General checks...".
In case of an error with this step (e.g. the Coveralls service is temporarily down) the job continues as normal and does NOT fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,11 +36,9 @@ jobs:
     - name: Perform build
       run: npm run build-test
 
-    # - name: Run unit tests
-    #   run: npm run coverage
+    - name: Run unit tests
+      run: npm run coverage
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -41,3 +41,4 @@ jobs:
 
     - name: Send report to Coveralls for package @ui5/linter
       uses: coverallsapp/github-action@v2.3.6
+      continue-on-error: true # Do not fail the job if coverage reporting fails (e.g. service is down)
```
d3xter666, line 44: This one seems to be wrong. What I can find in the documentation is `fail-on-error` property: https://github.com/coverallsapp/github-action
Can you share where this config is taken from?

PR #900, opened 2025-10-17
fix(rules/renderer): Detect missing apiVersion for method declarations

RandomByte:
Similar to the already supported property assigment with an arrow function, 'renderer' might also be declared as a method declarations or function declaration (which is transpiled to a method declaration).

If the 'renderer' property is set to a function/method, this implies apiVersion 1 and should therefore be reported as deprecated API usage.

Follow up of https://github.com/UI5/linter/pull/439

JIRA: CPOUI5FOUNDATION-878

Initially reported by @codeworrior

### DIFF
--- a/src/linter/ui5Types/SourceFileLinter.ts
+++ b/src/linter/ui5Types/SourceFileLinter.ts
@@ -340,6 +340,10 @@ export default class SourceFileLinter {
 				// i.e. { renderer: {apiVersion: 2, render: () => {}} }
 				this.analyzeControlRendererInternals(rendererMember.initializer);
 			}
+		} else if (ts.isMethodDeclaration(rendererMember)) {
+			// Analyze renderer property when it's a function (which gets transpiled to a method declaration)
+			// i.e. { renderer: function() => {} }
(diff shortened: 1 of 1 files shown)


test/fixtures/linter/rules/renderer/ControlRendererDeclaration_negative.js
```diff
@@ -19,24 +19,6 @@ sap.ui.define([
...
-		renderer: (oRm, oControl) => {
-		}
-	});
-
-	const NegativeExample6 = Control.extend("sap.ui.demo.linter.controls.NegativeExample6", {
-		metadata: {},
-		renderer(oRm, oControl) {
-		}
-	});
-
```
codeworrior: Not related to this change:
* line 43 (new): I think the comment should read "... But no renderer _has to be_ specified." I think we don't rule out extending the renderer for a view, but it's not mandatory to define one.
* line 66 (new): The comment states "... despite that Renderers are usually not present in the ts definitions...." Not sure whether this refers to the TypeScript types, then this is correct. If it refers to TypeScript sources:

PR #913, opened 2025-10-28
ci(package.json): Update node version to current LTS (22) and latest (24)

maxreichmann:
Follows up #912

`+` ci: Update node versions in GH Actions


RandomByte (changes requested):
This is a breaking change which can only be released in a new major version (v2) of UI5 linter. AFAIK we don't plan for that anytime soon

PR #990, opened 2026-01-28
fix: Teststarter false positives detection

d3xter666:
fixes: https://github.com/UI5/linter/issues/972


matz3:
Fixing false positives causes more warnings in our fixtures? Is that really the expected outcome? 🤔

d3xter666:
> Any update here? I should have explained my doubts more clearly. What surprised me is that there are now more messages for files where a message was already reported. To me this is unexpected and adds more false positives, as the concept can only be applied once per file, so I would expect only one message per file. Using the first script tag for the message seems reasonable to me (or whatever the current behaviour was).

Sorry, I've missed your comments!
I have now update the script so that it only reports one issue per file and points to the first script tag on the page

PR #1039, opened 2026-04-28
build(deps-dev): Upgrade knip from 5.88.1 to 6.7.0

werwolf2303:
JIRA: CPOUI5FOUNDATION-1222


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Use Node.js 20.11.0
       uses: actions/setup-node@v6
       with:
-        node-version: 20.11.0
```
RandomByte, line 22: Update the `name` attribute accordingly.

Also, could you please create a separate commit where you add `20.11.0` to the test matrix in https://github.com/UI5/linter/blob/main/.github/workflows/test.yml#L20C10-L20C19?

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false # Do not stop other jobs if one fails
       matrix:
-        version: [20, 22, 24]
+        version: ["20.19.0", 22, 24]
```
RandomByte: Should be `20.11.0` as per https://github.com/UI5/linter/blob/86882ea7aa47b0eb5161fcafb4ae4ac1d6cc7d69/package.json#L66

RandomByte (approved):

PR #1042, opened 2026-04-30
feat: Detect UI5 project specific globals

stopcoder:
Detects global access to project-namespace modules which bypass sap.ui.define/sap.ui.require. These were silently ignored by no-globals since app namespaces have no type definitions.
This new detection generates linter messages with `ruleId`s "no-project-globals".

JIRA: CPOUI5FOUNDATION-1242


src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1652,6 +1660,17 @@ export default class SourceFileLinter {
 					node,
 					fix: this.getGlobalFix(node),
 				});
+			} else if (this.#appNamespaceFirstSegment && ts.isIdentifier(exprNode) &&
```
codeworrior: `ts.isIdentifier(exprNode)` is already ensured by the outer "if"

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1652,6 +1660,17 @@ export default class SourceFileLinter {
...
+				exprNode.text === this.#appNamespaceFirstSegment) {
```
codeworrior: The code above handles cases where `globalThis` (or an alias of it) is used. The decision that the code takes is only reflected by `symbol`, not by `exprNode`. Means: the new branch doesn't to benefit from the above code and won't detect code like `globalThis.app.name.space`.

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1799,6 +1818,20 @@ export default class SourceFileLinter {
 			declarations.some((declaration) => checkFunction(declaration.getSourceFile()));
 	}
 
+	#isAppGlobalNotLocal(symbol: ts.Symbol): boolean {
```
codeworrior: Is there any similar code? Or what is the reasoning behind?

codeworrior (commented):
General question: I'M missing a documentation somewhere what the overall approach for the new check is.

From the code, my understanding is:
- assume that the current application exposes code globally (only) in that namespace
- extend the check for global to not only find globals from sap namespace, but also form this app namespace
- it does not detect globals from other namespaces (can for sure be extended to reuse libs etc.)

What the change does not: 
- it does not check that a global can be replaced by an AMD reference. It is up to the developer to ensure this
- it does not check whether global names map 1:1 to module names. Esp. in older apps that migrated from jquery.sap.require to sap.ui.define, this sometimes is not the case. This makes migration a bit harder. But I see such a check is not in scope for _this_ rule
- we know that there are SAP owned, third party-like libs that still use globals. They would have to be skipped by a generic "no globals" rule. But I guess by limiting the check to the app's namespace, it should be rather safe. Except, if app and external lib are not prefix-free.

src/linter/ui5Types/SourceFileLinter.ts
stopcoder: Done

src/linter/ui5Types/SourceFileLinter.ts
stopcoder: This is used for excluding the Local Variable use case, for example in the `test/fixtures/linter/rules/NoAppGlobals/NoAppGlobals_Negative.js`

If a user declares `const com = {}`, the TypeScript checker resolves `com` to a local symbol. The helper `#isAppGlobalNotLocal` prevents false positives.

stopcoder:
### Detection Strategy: Hybrid approach

The existing `no-globals` in JS uses TypeScript's type checker (symbol resolution). For app namespaces, we add **syntactic namespace matching** — checking if an identifier matches the project's known namespace prefix.

Both strategies coexist in `analyzePropertyAccessExpression()` as three branches:
- Branch 1 (existing): Type-checker resolves symbol → `isSymbolOfUi5OrThirdPartyType()` → report `NO_GLOBALS`
- Branch 2 (new): `exprNode.text` matches app namespace first segment → reconstruct full chain → verify prefix → report `NO_GLOBALS`
- Branch 3 (new): `isGlobalThis` access where `node.name` matches app namespace first segment → walk up chain → strip prefix → report `NO_GLOBALS`

### What Gets Detected

| Pattern | Detected? | Notes |
|---------|-----------|-------|
| `ns.Module.method()` in JS code | Yes | Branch 2: global method call |
| `var x = ns.Module` in JS | Yes | Branch 2: global read |
| `ns.Module = {...}` in JS | Yes | Branch 2: global assignment |
| `return ns.Module` in JS | Yes | Branch 2: global return |
| `window.ns.Module.method()` in JS | Yes | Branch 3: indirect global method call |
| `var x = window.ns.Module` in JS | Yes | Branch 3: indirect global read |
| `window.ns.Module = {...}` in JS | Yes | Branch 3: indirect global assignment |
| `"ns.Module"` in strings | No | String literal node, not expression |
| `// ns.Module` in comments | No | Not in AST |
| `ns.Module` where `ns` is local var | No | Local symbol found by `#isAppGlobalNotLocal` |
| `formatter="ns.Module.fn"` in XML | Yes | Already detected by `BindingLinter` as `no-globals` |
| `type="ns.Module.Type"` in XML binding | Yes | Already detected by `BindingLinter` as `no-globals` |
| `press="ns.Module.handler"` in XML event | Yes | Already detected by `Parser.ts` as `no-globals` |
| `factory="ns.Module.create"` in XML binding | Yes | Already detected by `BindingLinter` as `no-globals` |

src/linter/ui5Types/SourceFileLinter.ts
stopcoder: The globalThis case is now also covered in [66590bf](https://github.com/UI5/linter/pull/1042/commits/66590bfa5809fc3a7aa8ef889a441bf22a166db8)

d3xter666 (commented):
LGTM

Just one minor thing:
https://github.com/UI5/linter/pull/1042/changes#diff-28d3112b0317adccb934bf8c65a36bc55348ca0b6dcb70b00803ed600e7f51deR76 points to the cleaned up namespace, without the `window` prefix.
(https://github.com/UI5/linter/pull/1042/changes#diff-bcffcbfa2b6f09dc7cfe1b578e47c8193e8e1920ffacb431212297985ea6c2d5R10)

If we want to be consistent with `NO_GLOBALS` rule, the `window` must also be included:
- https://github.com/UI5/linter/blob/main/test/lib/linter/rules/snapshots/NoGlobals.ts.md?plain=1#L820

matz3 (commented):
LGTM.

+1 for the suggestion by @d3xter666 to align the `window` prefix handling for consistency.

matz3 (commented):
LGTM. UA review pending.

KlattG (approved):

matz3 (changes requested):
PR title doesn't fit to the changes

matz3 (approved):

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1652,6 +1661,33 @@ export default class SourceFileLinter {
...
+				const fullNamespace = extractNamespace(node as ts.PropertyAccessExpression);
+				if (fullNamespace.startsWith(this.#projectNamespaceDots + ".")) {
+					if (!symbol || this.#isProjectGlobalNotLocal(symbol)) {
+						this.#reporter.addMessage(MESSAGE.NO_PROJECT_GLOBALS, {
+							variableName: exprNode.text,
+							namespace: fullNamespace,
+						}, {node});
+					}
+				}
+			} else if (isGlobalThisAccess && this.#projectNamespaceFirstSegment &&
```
codeworrior, line 1675: Isn't this new check missing a check for the full namespace prefix? I only see terms that check against the `#projectNamespaceFirstSegment`, but no prefix check for the full namespace like in line 1667.

PR #1058, opened 2026-06-16
fix: Ignore allowed globals even when prefixed with a globalThis

### LINKED ISSUE #1057: UI5 linter 1.22.0 mistakenly complains about the use of allowed globals
## Expected Behavior

UI5 linter does not report the use of globals for the set of allowed globa,s even when they are prefixed by some expression that. evaluates to `globalThis`, e.g. `globalThis.sap.ui.loader.config`.

## Current Behavior

Since version 1.22.0, linter reports the use of a global.

## Steps to Reproduce the Issue

cd openui5/src/sap.ui.core
npx @ui5/linter@1.22.0 test/sap/ui/core/qunit/loader/amd/

Observe the findings for `.../amd/base_tag/_launchTest.js` and others.

## Context

- UI5 linter version: **`1.22.0`**

## Other Information

See code comment in [PR 1042](https://github.com/UI5/linter/pull/1042#discussion_r3421745859).

codeworrior:
Fixes #1057


codeworrior:
Tests are still missing!

test/fixtures/linter/rules/NoGlobals/NoGlobals.js
```diff
@@ -1,4 +1,4 @@
-sap.ui.define(["sap/ui/core/mvc/Controller"],
+sap.ui.define(["sap/ui/core/mvc/Controller", "sap/ui/test/Opa5"],
```
codeworrior: Oops, this change needs to be reverted. It's a leftover from some experiments.

In theory, the linter should also not complain about `Opa5.getWindow().sap.ui.require`. I tried this but it suffers from the same problem as indirect global access via `top` or a top-level `this`. The type for all those nodes is not detected as a `globalThis` equivalent (which, BTW, is understandable as TypeScript IMO cannot know whether the "other" window has the same APIs available as in the project's scope).

src/linter/ui5Types/SourceFileLinter.ts
```diff
@@ -1653,7 +1653,7 @@ export default class SourceFileLinter {
 			if (symbol && this.isSymbolOfUi5OrThirdPartyType(symbol) &&
 				!((ts.isPropertyAccessExpression(node) || ts.isElementAccessExpression(node)) &&
 					this.isAllowedPropertyAccess(node, isGlobalThisAccess))) {
-				const namespace = extractNamespace((node as ts.PropertyAccessExpression));
+				const namespace = extractNamespace((node));
```
codeworrior: unnecessary double parentheses: ((node))

d3xter666: Fixed with https://github.com/UI5/linter/pull/1058/commits/1641a536978607d53a722ec02c57113320c002ce

test/fixtures/linter/rules/NoGlobals/NoGlobals.js
codeworrior: Import removed.

Team consent is to accept the false-positive reports of `Opa5.getWindow().sap.ui.require` calls for now.

test/lib/linter/_linterHelper.ts
```diff
@@ -143,13 +133,32 @@ export function createTestsForFixtures(fixturesPath: string, fix = false) {
...
-					testName: fileName,
-					fileName,
-					fixturesPath,
-					filePaths: [fileName],
-					fix,
-				});
+				let namespaces: (string | undefined)[] = [undefined];
+				if (fixturesPath.includes("NoProjectGlobals") ||
+					fixturesPath.includes("NoGlobals")) {
+					// Tests for NoProjectGlobals and NoGlobals rules should be tested with different namespaces to
```
matz3: Do we really need to also run the NoGlobals test in this matrix?
Is there an expected difference of the results? If not, then I would propose that we reduce the snapshot size by assuming that all 3 namespaces yield the same result, and only including that into the snapshots.

codeworrior: I had proposed this. But now, as the different cases for implicit global access of allowed APIs became part of the `NoProjectGlobals.js`, I'm fine with omitting the `NoGlobals.js`

matz3: Done

matz3 (approved):

test/lib/linter/_linterHelper.ts
codeworrior: Thanks!

flovogt (approved):

PR #1126, opened 2026-08-26
feat: Don't enforce a renderer for subclasses of HTMLElement

codeworrior:
UI5 runtime introduces a new base class for native HTML controls and webcomponents. This base class implements renderer inheritance. A missing renderer property in a subclass therefore should not be reported as an error.


RandomByte (approved):
LGTM

